PR #1, opened 2020-04-02
Attempt to stand up a WFIRST SIT cluster

cslocum:
This is mostly for working through bugs, and plenty have been encountered.

I will start documenting them here (should have been doing that all along...)


yuvipanda:
I'm going to merge this now, mostly so staging represents what we're playing with closer than it does now. Awesome work! \o/

cslocum:
Thanks Yuvi!

PR #3, opened 2020-05-11
[WIP] Use CircleCI for public untrusted CI checks on PRs

yuvipanda:
GitHub actions have a disk space limit of 14G. Our images
do not fit into that. CircleCI does not have this limitation,
so let's switch to that.

While deployment that requires access to credentials will
happen with space telescope internal Jenkins, CircleCI can
be used to run external *untrusted* code. Since anyone can
make a PR, it'll be more secure to run this CI in an
external system that has no access to our AWS infrastructure

PR #4, opened 2020-05-11
Add wfirst-sit deployment

cslocum:
TODO: description


.gitignore
```diff
@@ -104,4 +104,6 @@ venv.bak/
 .mypy_cache/
 
 # fetched helm charts
-hub/charts/
\ No newline at end of file
+hub/charts/
+
+deployments/*/secrets/*
```
yuvipanda, line 109: Nice!

deployments/wfirst-sit-dev/hubploy.yaml
```diff
@@ -0,0 +1,17 @@
+images:
```
yuvipanda, line 1: I'd like to understand what the purpose of the '-dev' deployment would be. Is it for individual people to hack on? Experimental tests to wfirst-sit that may or may not make it to prod? 

The problem with this is that things will drift apart between wfirst-sit and wfirst-sit-dev. So things that work in wfirst-sit-dev will not work in wfirst-sit, and vice versa. So depending on the reason for having a separate -dev deployment, let's see what is a good solution for it? It might be another 'dev' branch, a way for people to test this locally, deploy from your own fork, etc. But I'd like to avoid having '-dev' variants of each hub, for the drift reasons.

yuvipanda: This drift is already happening - for example, https://github.com/spacetelescope/jupyterhub-deploy/pull/3 added a `WORKDIR` to the Dockerfile, which is missing here.

PR #8, opened 2020-06-16
Minor refinements to wfirst-sit

cslocum:
- setup dummy auth, with admin and whitelist
- add vim, emacs, and gfortran


cslocum:
Mike says it's OK to merge.

PR #9, opened 2020-07-15
First wack at the ROMAN SIT deployment

cslocum:
There are definitely still some issues to work out here.  For instance, the Dockerfile needs attention.

But it would be good to get these changes in, so that the fine tuning can be seen independently.


cslocum:
Mike looked at this and approved it for merging.  So I will do that.

PR #10, opened 2020-07-28
[WIP] Document new JupyterHub deployment process

cslocum:
See https://jira.stsci.edu/browse/JUSI-423


cslocum:
TODO: note that docker service needs to be running

PR #11, opened 2020-08-05
Documented voodoo for .sops.yaml and trust relationships

jaytmiller:

PR #12, opened 2020-08-11
Misc. doc updates

cslocum:
This PR includes the changes Todd made in https://github.com/spacetelescope/jupyterhub-deploy/pull/11.


doc/example-sops.yaml
```diff
@@ -0,0 +1,4 @@
+creation_rules:
+  - path_regex: .*
+    kms: "arn:aws:kms:us-east-1:162808325377:key/5f56d94b-a658-4c92-a226-58bf07c88890"
```
TheRealZoidberg, line 3: Should we have an actual key here?

doc/example-common.yaml
```diff
@@ -26,30 +26,33 @@ jupyterhub:
   hub:
     extraEnv:
       OAUTH2_AUTHORIZE_URL: "https://auth.mast.stsci.edu/oauth/authorize"
-      OAUTH_CALLBACK_URL: "https://roman-sit.science.stsci.edu/hub/oauth_callback"
+      # OAUTH2_AUTHORIZE_URL: "https://auth.masttest.stsci.edu/oauth/authorize"
```
TheRealZoidberg, line 29: Should we have the actual URL here? Could make it a target

TheRealZoidberg: Same with url's below

cslocum: Can you clarify please?  Not sure what you mean.

doc/example-sops.yaml
cslocum: Not sure.  The ARN does not reveal anything about the key.  The key is secret, not the ARN

doc/DEPLOYMENT.md
```diff
@@ -198,11 +199,11 @@ Since we use sops to encrypt and decrypt the secret files, we need to copy the *
...
 
-Due to a hiccup documented in [JUSI-412](https://jira.stsci.edu/browse/JUSI-412) it is necessary to manually insert the ARN of the decrypt role into *staging.yaml* so that sops can decrypt the file during deployment without specifying the role.  After *staging.yaml* has been created and configured, edit the file (**do not use sops**) and add the role ARN.  You can see an example of the bottom of the an updated, encrypted file [here](https://github.com/cslocum/jupyterhub-deploy/blob/staging/doc/example-staging-with-inserted-role.yaml).
+**BUG**: After *staging.yaml* has been created and configured, sops adds a section to the end of the file that defines the KMS key ARN and other values necessary for decryption.  Due to a hiccup documented in [JUSI-412](https://jira.stsci.edu/browse/JUSI-412), it is necessary to manually insert the ARN of the decrypt role into the file so that sops can decrypt the file during deployment without specifying the role.  Edit the file (**do not use sops**) and add the role ARN.  You can see an example of the bottom of the an updated, encrypted file [here](https://github.com/cslocum/jupyterhub-deploy/blob/staging/doc/example-staging-encrypted.yaml).
```
TheRealZoidberg, line 206: reference to cslocum repo

doc/example-sops.yaml
cslocum: Also, I like to keep the general format/example of something similar in these files so people can more easily understand what should go in there.

doc/DEPLOYMENT.md
cslocum: Yep - as I mentioned in the PR description, the last thing before merging is to replace "cslocum" with "spacetelescope"

doc/example-common.yaml
TheRealZoidberg: https://roman-sit.science.stsci.edu/hub/oauth_callback url - by showing this here it could make it a target for a DDOS or some sort of attack.  Not sure if I'm being overly cautious

doc/example-sops.yaml
TheRealZoidberg: Right, makes sense.

doc/example-common.yaml
cslocum: Hmm.  That is a good point.  Although, that URL is useless if the credentials are unknown.

cslocum: I would think DDOS attacks are MAST's responsibility to handle.

doc/DEPLOYMENT.md
```diff
@@ -136,50 +156,54 @@ To get started, clone the repository: `git clone https://github.com/spacetelesco
...
 
-A file named *common.yaml* file needs to be created in the *config* directory.  An example can be found [here](https://github.com/cslocum/jupyterhub-deploy/blob/roman/doc/example-common.yaml).  Place a copy of this example file in *config*, and edit the contents as appropriate.
+3) A file named *common.yaml* file needs to be created in the *config* directory.  An example can be found [here](https://github.com/cslocum/jupyterhub-deploy/blob/staging/doc/example-common.yaml).  Place a copy of this example file in *config*, and edit the contents as appropriate.
 
-Add, commit, and push all changes.
+4) Add, commit, and push all changes.
 
-Once the configuration changes have been made, change directories to the top level of the jupyterhub-deploy repository.  Then issue this command to build the Docker image and push it to ECR: `hubploy build <deployment-name> --push --check-registry`.
+5) From the top level of the jupyterhub-deploy repository, issue this command to build the Docker image and push it to ECR: `hubploy build <deployment-name> --push --check-registry`.
```
jaytmiller, line 167: IMHO,  and this applies everywhere,  EVERY command should be on a separate line using the "-" notation to start if off,  or else a ``` block also on separate lines.   I think this is really important,  because part of executing this stuff going through and cutting and  pasting...  which is harder mid-sentence,  and possibly with line wraps.     E.g.  check this out:  https://github.com/spacetelescope/jwst,  or this: https://github.com/astroconda/astroconda/blob/master/source/releases.rst

cslocum: Sure, I can do that.  Makes sense.  About to meet with Joe C in a few mins, but I'll fix this after.

jaytmiller:
I noticed one thing which got dropped from #11:  the trust relationships.   That's one of the more complex manual tasks I have to do for every test iteration.  Are those going to carry forward?   If not,  can we make a file of hints which *dont* carry forward?

jaytmiller:
Another thing got dropped:  the ECR login mantra.  (I don't care how that  gets expressed or have opinions on what is correct,  I just want something I can cut-and-paste by rote.   That worked for me.)

cslocum:
Thanks Todd.  I was hoping to leave out the trust relationships because going forward, we need to eliminate that step.  That's part of what I was discussing yesterday about needing the master role with the perm boundaries attached.

I will look at the ECR login mantra.

PR #13, opened 2020-09-14
small edits after trying instructions

super-cob:


cslocum:
@super-cob I think this can be closed (not merged).  I pulled in your changes that made sense, while a few of them are now obsolete.  Do you agree?

PR #18, opened 2020-10-22
"roman" deployment for dev,  resolves docker build SSL issues

jaytmiller:
This builds with a warning installing: 

`jupyter labextension install @jupyterlab/server-proxy`

of

`Config option 'kernel_spec_manager_class' not recognized by 'InstallLabExtensionApp'`

But I think pip is generally working,  all the environments build and e.g. "import jwst" works in roman.


jaytmiller:
I attached diffs relative to my original copy from roman-sit on this branch.

[dev.diffs.txt](https://github.com/spacetelescope/jupyterhub-deploy/files/5426236/dev.diffs.txt)

cslocum:
Todd, this looks good to me.  Are you done making/testing changes?

jaytmiller:
This is "done enough" to merge.   A lot of this stuff is subject to change though as we sort out CI/CD and testing in general.   For now "image-build" is how you build the image easily,  basically instantiate the template and image-build.

jaytmiller:
One other thing:  this is still set up for "roman" vs. "roman-sit".

cslocum:
Well, what I just setup on the dev account is "dev.roman", but I think what you have in here is fine.  We'll always be changing the name depending on what we're deploying.  It's easy enough to copy what you have to roman-dev (or whatever the directory gets named) and rename files/replace strings in files.  That's pretty much standard procedure.

One thought I had was that we will need to update the docs to support your new scripts (which are awesome BTW).  But I think we can tackle that after merge.  I'm thinking that we can merge this and my (shortly coming doc PR) without reconciling them, and then do the teardown/re-deploy of dev.roman that I intend for as early as tomorrow, updating the docs as I go.

What do you think?

jaytmiller:
I'm fine with merging and think it's necessary for pushing on to TEST and PROD.   I'll whip up some docs for the scripts.   These are subject to change as we dig deeper into testing and CI/CD but for now I think we need at least "image-build" to get through certs.

jaytmiller:
This PR was re-issued against the "integration" branch vs. the already merged "integration-doc" branch.

PR #22, opened 2020-11-02
Sample configs: revert to manual proxy

cslocum:
Since we are no longer offloading SSL handling to ACM.


jaytmiller (approved):
LGTM,  we should continue batching changes in "integration" and then do a test drive.

PR #25, opened 2020-11-12
Script updates to fix deploy-all,  adds prototype secrets-xxx scripts.

jaytmiller:
NOTE:  image-test is broken in this version,  it's a work-in-progress which is affected by the common base image, etc.


cslocum (approved):
Awesome!

PR #27, opened 2020-11-18
Add common base image,  testing approach

jaytmiller:
Added common base image based on jupyter/scipy-notebook
Added scheme for building environments.
Added semi-automatic scheme for updating pinned environment specs via "floating" and "frozen" build modes.
Added baseline roman image derived from common image,  supports cvt, mirage, and roman-cal kernels.  Has sample tests.
Added scheme for import and notebook testing on a per-environment basis,  generalized image-test entrypoint.
Added prototype tess image,  still very rough,  no frozen build successful yet but based on common image.
See doc/FRAMEWORK.md for more explanation.


deployments/common/image/common-scripts/env-clone
```diff
@@ -0,0 +1,17 @@
...
+
+. /etc/profile.d/conda.sh
+
+if [[ "$USE_FROZEN" == "1" ]]; then
+    /opt/common-scripts/env-create  $to  /opt/environments/${to}/${to}-frozen.yml
+    exit $?
+fi
+
+echo "-------------------------------------------------------------------------------"
+echo "Cloning env $to from env $1"
```
cslocum, line 16: Should probably be $to here instead of $1.

I also don't see $from used anywhere in this script.

deployments/jmiller-hub/config/common.yaml
```diff
@@ -1,13 +1,45 @@
...
+      - cslocum@stsci.edu
+      - jmiller@stsci.edu
+      - gough@stsci.edu
+      - jmatuskey@stsci.edu
+    admin:
+      users:
+      - cslocum@stsci.edu
+      - jmiller@stsci.edu
+      - gough@stsci.edu
+      - jmatuskey@stsci.edu
```
cslocum, line 45: We probably shouldn't have our emails in here, since this is public.  Maybe put a placeholder in the example file in docs/ ?

deployments/roman/image/environments/cvt/cvt-frozen.yml
```diff
@@ -0,0 +1,303 @@
...
+    - sphinxcontrib-jsmath==1.0.1
+    - sphinxcontrib-qthelp==1.0.3
+    - sphinxcontrib-serializinghtml==1.1.4
+    - stsci-rtd-theme==0.0.2
+    - tenacity==6.2.0
+    - textwrap3==0.9.2
+    - toml==0.10.2
+    - typed-ast==1.4.1
+    - yarl==1.6.3
+prefix: /opt/conda/envs/cvt
```
cslocum, line 303: What does "prefix" do?

jaytmiller: prefix is defining where that environment  lives in conda

deployments/jmiller-hub/config/common.yaml
jaytmiller: I  just neutered  these in jmiller-hub.   The template file already has this.

deployments/common/image/common-scripts/env-clone
jaytmiller: $from is in there,  I switched  to it.

PR #32, opened 2020-11-23
Automatic ECR scan developer report on push.

jaytmiller:
Added wait for ECR scan,  automatic report and summary to image-push.

This  both captures the AWS scan output as a YAML file and follows up by checking CVE status with Ubuntu to see if there are any fixes available,   also added to the YAML.

Refactored setup-env.template into setup-env.template and derived-env to make it simpler to configure and add to environment without forcing dev env updates.   The part unlikely to change is in derived-env.

Also added "personal build" mode for e.g. doing laptop builds by setting PERSONAL_IMAGE=1.  Untested.

Submitting for next sprint integration.

### DIFF
--- a/derived-env
+++ b/derived-env
@@ -5,14 +5,10 @@
 
 if [[ "$PERSONAL_IMAGE" == "1" ]]; then
     # Standalone JupyterLab setup,  no ECR scanning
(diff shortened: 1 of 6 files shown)


cslocum (approved):
Looks great!

PR #33, opened 2020-11-24
Fix conda certs

jaytmiller:
Mostly adds /opt/common-scripts/fix-certs to search conda for openssl certs and replace them with ITSD's firewall packet inspection enabled cert for DEV, TEST, OPS.   
Added convenience script 'image-root-sh' to handle the common debug case of getting root in a running debug image w/o sudo or explicitly reconfiguring the environment.
Added "time"  to image-test so we develop a sense of  how long unit tests should take to run in each image.  
Set CRDS_VERBOSITY=20 which should show download progress while the test notebook or any work does CRDS downloads,  3.6G for the tests is too long a pause with no output.


cslocum (approved):
This is great!  Too bad it's necessary...but this looks good to me.

PR #34, opened 2020-12-01
Add ds9 rebased

jaytmiller:
This PR adds DS9 as a separate environment in the  common base image.
Also adds docker-prune, docker-wipe, run-lab convenience scripts.


cslocum (approved):
This looks great!  Thanks.

The COPY "/" or no "/" is always a gotcha for me.

PR #35, opened 2020-12-02
Increase memory allocation in roman deployment.

jaytmiller:
Updated memory allocations in roman and jmiller-hub deployments to 8G guaranteed 15G limit so JWST cal tests (standing in for Roman) pass.


cslocum (approved):
Looks good to me!

PR #36, opened 2020-12-02
make sure user storage is from EFS

super-cob:
Foundational to our setup but overlooked - this is crucial for more than just AMI rotation, also users need to be able to be scheduled on the correct nodes.


cslocum (approved):
:+1:

PR #37, opened 2020-12-04
Clean up kernel init code

jaytmiller:
Removed ineffective redundant jupyterhub kernel initialization,  automatic conda kernels should be OK for now.


cslocum (approved):
Looks good to me!

PR #38, opened 2020-12-08
awsudo sops in deploy instead of updating encrypted yaml file

super-cob:
in my experience you could get thru the deployment without modifying the encrypted yaml to include ARN if you did an awsudo sops on the file every time. much easier to script so it's a permanent change i'm proposing.

My intention was to remove every reference to encrypted sops modification in docs as well as make the script change, in case i missed something


cslocum:
https://github.com/spacetelescope/jupyterhub-deploy/blob/main/doc/example-sops.yaml should be deleted as well.  Otherwise, :+1: .

cslocum (approved):

PR #43, opened 2021-01-07
Update secrets script for new dir location

jaytmiller:
Updated secrets-edit and secrets-push for change in location of secrets directory.


jaytmiller:
Now included in TESS DEV setup PR.  Withdrawing this copy.

PR #44, opened 2021-01-12
Tess dev hub setup

jaytmiller:
Fixes secrets scripts broken by move of secrets directory.
Moves e-mails from config/common.yaml to the secrets template.
Sets TESS memory allocations to 4G guaranteed, 10G limit basically assuming t3.xlarge.
Adds new EFS configuration to TESS common.yaml,  roman still needs it.
Adds ssh session timeout and Terraform logging env variables to setup-env.template.
Update TESS EFS configuration in common.yaml to avoid creation of EBS,  node affinity conflicts
Switched image tag from <mission>-latest back to latest in TESS, Roman, setup-env.template.

### DIFF
--- a/deployments/roman/config/common.yaml
+++ b/deployments/roman/config/common.yaml
@@ -35,9 +35,3 @@ jupyterhub:
     memory:
       guarantee: 8G
       limit: 15G
-  auth:
-    users:
-      - add-me@stsci.edu
-    admin:
-      users:
-         - add-me@stsci.edu
--- a/deployments/tess/config/common.yaml
+++ b/deployments/tess/config/common.yaml
@@ -17,7 +17,7 @@ jupyterhub:
       enabled: true
   singleuser:
     image:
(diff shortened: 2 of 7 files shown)


deployments/roman/config/common.yaml
```diff
@@ -35,9 +35,3 @@ jupyterhub:
     memory:
       guarantee: 8G
       limit: 15G
-  auth:
```
super-cob, line 38: can we put this in secrets instead?

jaytmiller: Yep,  that's what I did.  I removed it from common.yaml and added it to the secrets template.  It worked fine.  I removed it from common.yaml for both tess and roman.

jaytmiller: (It worked fine with real e-mails...)

jaytmiller:
Resolved all comments with Jake, merging.

PR #46, opened 2021-01-22
Functionality to add landing page user documentation

cslocum:
- extend the spawn.html template to allow for added text
- set the cull time for inactive user notebook servers to 30 mins (default is 60 mins)

Note: the text in spawn.html is just placeholder text.  Future PR(s) will update the content.


jaytmiller (approved):
Over my head,  but looks clean to me and looked good in action.

PR #47, opened 2021-01-28
rename tess to tike

super-cob:
rename the project from tess to tike means just changing the folder, but that kind of renames all the files within


cslocum (approved):
This looks good to me.  Thanks.  Merge away!

PR #49, opened 2021-01-29
here's new proper example configs also applied to roman test

super-cob:
Here you find some better configs in the examples - also applied to roman test. Gratuitous --debug flag thrown in to deploy script which i find helpful.


cslocum (approved):
Looks good to me!  Thanks.

PR #50, opened 2021-01-29
Hide ~/.kube/config after update

jaytmiller:
After updating kubeconfig in deploy-all,  chmod it 600 to prevent helm complaints and/or malfunction.


cslocum (approved):
This looks good to me!  Thanks Todd.

PR #52, opened 2021-02-19
update jupyterhub helm chart to 0.11.1 which is 1.3.0 jh version

super-cob:
update helm chart and jh version with a restructured config and include pdb for proxy which was needed


cslocum (approved):
Nice!  I like that you've pulled in all put a small bit of singleuser out into it's own file.

This looks good to me!

PR #53, opened 2021-02-19
Implement gitpuller for tike_content repo

cslocum:
Also, the example-notebook content was removed.


TheRealZoidberg (approved):
Looks good to me.

jaytmiller (approved):
LGTM but I'm wondering how this affects testing...  and  have some ideas.   

Each deployment has an arbitrary environments/test script which is what image-test runs;  currently that just calls a generic script which runs notebooks and import tests defined for each environment.  The top level test script could do the same rsync,  then add example notebooks into /opt/environments/... where the tike tests are located.   Then  image-test would run them,  and they'd also update to the latest version whenever tike_content  changes.  They would not be permanent,  they'd only exist as long as the test container does.   Alternately we could build tike_content into the Dockerfile so that some version of them exists in the image and is tested.   Either way it's probably a follow-on ticket.   One last thing to note is that the tike environment already inherited a lengthy list of notebook tests and example notebooks which are installed by the Dockerfile and that should either be made to work or dropped as the definition of "working tike image" changes.

cslocum:
Sorry Todd!  I only just saw this, my github account is linked to gmail, and this wound up in a folder I rarely check.

Seems easiest to put the tike_content into the container for testing purposes.  It's redundant from the standpoint of it will get populated via the post start hook anyway, but there are really no downsides to cloning the content in the image for testing purposes.

I have also been lax with updating the test scripts.  I need to go back and do that.  Sorry I missed your comment before merging.

jaytmiller:
No problem, I can't keep up with all the e-traffic either.

PR #54, opened 2021-02-23
example secrets with new layout

super-cob:


jaytmiller (approved):
LGTM,  should have merged this a while ago.

PR #58, opened 2021-03-25
Kernels and tests

jaytmiller:
Mod to test-notebooks to change directory to same as test notebook to pick up relative access to included test files like pipeline  configs.
Created  /opt/environments/setup-notebooks called in both common.yaml and in Dockerfile.custom (for now).   It's an entry-point which does whatever is needed,  currently nbgitpuller and set-notebook-kernels as appropriate.   This entry-point  makes notebook setup the  same in standalone mode and hub mode.  Eventually all missions should do whatever they need to do.
Wiped jwebbinar/tests since they're all TESS/Tike tests and don't pass.
Re-implemented  docker-wipe script as using "docker system prune --all --force" since  it clears all images, containers,  and the build context which I found was persisting even after all built containers and images were deleted.


deployments/jwebbinar/config/common.yaml
```diff
@@ -1,4 +1,4 @@
-# Use this  CRDS config,  maybe with a different S3 bucket, for scalable performance on AWS.
+ Use this  CRDS config,  maybe with a different S3 bucket, for scalable performance on AWS.
```
cslocum, line 1: I think this line should be commented?

jaytmiller: Definitely.  No idea how that happened.

cslocum (approved):
This all looks great Todd, thank you!

PR #61, opened 2021-03-29
Reduce CRDS verbosity

jaytmiller:
Unset CRDS_VERBOSITY setting to reduce verbosity,  added comment about possible settings.


cslocum:
Thanks!  Looks good to me.

PR #62, opened 2021-03-30
Fix frozen builds

jaytmiller:
Added missing "COPY environments/frozen ..." to roman and tike deployments.

NOTE:  even with this fix,  the roman frozen build failed for me due to dependency conflicts I haven't sorted out.   Nevertheless, w/o this fix,  the roman frozen build would fail regardless of what  the specs are.

### DIFF
--- a/deployments/roman/image/Dockerfile
+++ b/deployments/roman/image/Dockerfile
@@ -27,6 +27,9 @@ ENV CRDS_VERBOSITY=20
 
 USER ${NB_UID}
 
+# All specs for frozen builds need to be available before their normal installs
+COPY environments/frozen/  /opt/environments/frozen
+
 # Roman CAL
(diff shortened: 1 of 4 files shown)


deployments/roman/image/Dockerfile
```diff
@@ -107,7 +104,20 @@ RUN apt-get update && \
...
 
+# For standalone operation outside JupyterHub,  note that  /etc also includes
+# common home directory files.
+
+RUN tree -a /etc/default-home-contents/
+
+RUN cp -r /etc/default-home-contents/* /home/jovyan && \
+    fix-permissions  /home/jovyan
+
+RUN tree -a /home/jovyan
```
cslocum, line 115: These `tree` commands will be removed in a coming PR, but this is fine.

cslocum (approved):
This looks good.

PR #64, opened 2021-04-01
Tweaks to fix tike and roman builds and testing

jaytmiller:
Modify tike and roman Dockerfiles to chown anything copied to /opt to NB_UID:NB_GID.  This mimics scipy-notebook and also enables chdir'ing to test directories in /opt that supply supporting files to test notebooks and also supports writing outputs in the CWD.   This writability used to be accomplished  by a very late fix-permissions but that was dropped because it invalidated the Docker cache for all environments.

Added default-home-contents to tike to make Dockerfile.trailer work,  also sets up astroquery for MAST.  

Switched roman-cal from a git checkout to PyPi romancal-0.2.0.

Testing is "fixed" in the sense that tests don't fail because the framework isn't set up right,  they fail because they fail.

tike and roman frozen builds are obsolete,   their updates didn't build so were not committed.

### DIFF
--- a/deployments/common/image/Dockerfile.trailer
+++ b/deployments/common/image/Dockerfile.trailer
@@ -25,7 +25,7 @@ RUN /opt/common-scripts/fix-certs
 
 # This top level command runs unit tests for the image,  nominally import tests
 # and notebook runs but it's an arbitrary bash script.
-COPY environments/test    /opt/environments/test
+COPY --chown=${NB_UID}:${NB_GID} environments/test    /opt/environments/test
 
 # Install security patches now that build is complete;  potentially this could be done sooner
 RUN apt-get update && \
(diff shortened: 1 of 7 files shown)


deployments/common/image/Dockerfile.trailer
```diff
@@ -25,7 +25,7 @@ RUN /opt/common-scripts/fix-certs
 
 # This top level command runs unit tests for the image,  nominally import tests
 # and notebook runs but it's an arbitrary bash script.
-COPY environments/test    /opt/environments/test
+COPY --chown=${NB_UID}:${NB_GID} environments/test    /opt/environments/test
```
cslocum, line 28: Cool!  I didn't know you could do this.

deployments/tike/image/Dockerfile
```diff
@@ -89,13 +89,9 @@ RUN apt-get update && \
 # For standalone operation outside JupyterHub,  note that  /etc also includes
 # common home directory files.
 
-RUN tree -a /etc/default-home-contents/
```
cslocum, line 92: Sorry - I thought these were already removed in main.

cslocum (approved):
This all looks great, Todd.  Thanks!

PR #68, opened 2021-04-13
Display MD file on startup; render MD by default

cslocum:
Also upgraded to JupyterLab v3.

Addresses https://jira.stsci.edu/browse/JUSI-607.


deployments/tike/config/common.yaml
```diff
@@ -52,7 +52,9 @@ jupyterhub:
             - "-c"
             - >
               rsync -r /etc/default-home-contents/ /home/jovyan;
-              gitpuller https://github.com/spacetelescope/tike_content master /home/jovyan/tike_content
+              gitpuller https://github.com/spacetelescope/tike_content master /home/jovyan/tike_content;
```
jaytmiller, line 55: We should start a setup-notebooks for tike so that this stuff can be installed in the standalone image as well.   setup-notebooks may really be too narrow a name,   it's more like setup-home.    And I think maybe  ALL of that goes into setup-home,  including the rsync,  because for the standalone image it only gets called when the image is built.   I'm fine with merging this but think we should continue evolving setup-notebooks/setup-home.

jaytmiller (approved):
One comment,  LGTM

deployments/tike/config/common.yaml
cslocum: You're right, that's a good idea.  I might do that now.  I like `setup-home`.  Not having things in the Docker image as well also impact security scanning.

jaytmiller: Nice point about security!

jaytmiller (approved):
LGTM,  more-or-less what I was thinking.   One other frame-worky thing occurred to me:  we should start thinking about generalizing the startup markdown the same we we're doing global_bashrc and home setup.   Each deployment should have it's own markdown file which eventually can get handled by the multi-mission config yaml (all.yaml?)

PR #71, opened 2021-04-20
Implement source code and image scanning

jaytmiller:
Added source code scanning with bandit, safety, and Anchore.
Revised env-* installation scripts to optionally use mamba vs. conda by setting $CONDA_VER.
Implemented baseline sscan-bandit and sscan-find-python-files scripts.
Implemented baseline sscan-safety and sscan-find-frozen-specs scripts.
Implemented developer script sscan to run both sscan-bandit and sscan-safety.
Fixed security issues with Python tools detected by bandit.
Modify test-notebooks to glob inputs and improve temp directory handling.   Fix jwebbinar notebook glob.
Modify image-freeze to capture common ds9 environment.
Updated frozen specs.
Remove .circleci
Update astroquery.cfg


cslocum:
> Revised env-* installation scripts to optionally use mamba vs. conda by setting $CONDA_VER.

What's the reasoning behind this?

.github/workflows/source-code-scanning.yml
```diff
@@ -0,0 +1,59 @@
...
+        USE_FROZEN: [ 0 ]
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
cslocum, line 31: I think in some cases (jwebbinar?) we are not using Python 3.8, for package compatibility issues.  Even if that's not the case, the implication here is that the test wouldn't be valid if a deployment uses a different Python version, right?

setup-env.template
```diff
@@ -23,6 +23,11 @@ export USE_FROZEN=1
 # Set to 1 for e.g. personal laptop build
 export PERSONAL_IMAGE=0
 
+# Select "conda" tool used to install conda package and environments: conda or mamba
+#   mamba is largely a drop-in replacement for mamba with improved dependency resolution,
```
cslocum, line 27: > mamba is largely a drop-in replacement for mamba

Do you mean "mamba is largely a drop-in replacement for **conda**">

jaytmiller:
The reasoning behind CONDA_VER is to optionally enable using mamba instead of conda since it is supposedly faster/better.  Currently conda is used and small changes will need to be made to image-build to pass CONDA_VER into docker builds.   When mamba was used by default as I developed,  I think it shaved 12 minutes off the build time,  1 hr 14 min vs 1 hr 26 minutes for the build, test, scan workflow.

cslocum:
Do we need to update user data for the CI node to install bandit and safety?

cslocum:
> When mamba was used by default as I developed, I think it shaved 12 minutes off the build time, 1 hr 14 min vs 1 hr 26 minutes for the build, test, scan workflow.

Nice!

cslocum:
This is fantastic Todd!  Nice work.

.github/workflows/source-code-scanning.yml
jaytmiller: That particular Python-3.8 is just defining the environment which runs bandit and safety.  It is not directly related to environments we install in our images.

PR #73, opened 2021-04-22
Build scipy-notebook base image from source

cslocum:
Addresses https://jira.stsci.edu/browse/JUSI-644

This introduces changes to how we build our Docker images.  We are no longer pulling anything from Docker Hub, instead we are cloning https://github.com/jupyter/docker-stacks and modifying the make args in the build process.

The general process is:

1) Build minimal image from Ubuntu 20.04, installing and configuring certs on top of it
2) Clone docker-stacks and build base-notebook with overridden ROOT_CONTAINER (with what was built in step 1)
3) Build minimal-notebook, then scipy-notebook, on top of base-notebook
4) Build our common + deployment specific image on top of scipy-notebook

I will mention that the image size increased from 5.3 GB to 7.5 GB, which is not great.

NOTE: we're going to hold off on merging this until we have a break in between JWebbinar sessions.


deployments/common/image/Dockerfile
```diff
@@ -102,40 +99,29 @@ RUN wget https://sourceforge.net/projects/turbovnc/files/2.2.6/turbovnc_2.2.6_am
...
-# BUG: remove this!!!
-RUN npm config set strict-ssl false
+# This must be run after conda installs
+RUN npm config -g set cafile /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
 
+RUN git config --global http.sslCAinfo /etc/ssl/certs/ca-certificates.crt
+
+# ------------------------------------------------------------------------
 USER ${NB_UID}
+
```
jaytmiller, line 115: Not critical,  but this one should probably be deleted since the above is global.

jaytmiller (approved):
Sweet!

deployments/common/image/Dockerfile
cslocum: Done.

PR #74, opened 2021-04-23
Fix secrets-get-exports broken by switch to literal_eval()

jaytmiller:


jaytmiller (commented):
Verified this twice,  not possible to break it more than it is now,  merging.

PR #75, opened 2021-04-28
Fix image-update

jaytmiller:
Fix missing "!" in the image-update shebang theoretically missing since October.  Not sure why this has been working well enough nobody noticed...  but on OS-X there are at least some times it silently fails and does nothing.


jaytmiller:
Since this is a one character change I just tested,  self-merging.   This may only affect OS-X.

PR #76, opened 2021-04-28
Add image-patch

jaytmiller:
Removed apt-get dist-upgrade from Dockerfile.trailer since the need for
  it is offset by (a) building scipy-notebook from source ourselves and
  (b) supporting patch images which can *optionally* do the same thing.
Updated mission Dockerfiles for new Dockerfile.trailer w/o apt-get
dist-upgrade.
Added basic image-patch and image-pull scripts.
Modified image-push anticipating ITSD quarantine scheme.
Updated CI actions to run on PR or push/merge to main branch only.
Updated CI periodic run to weekly (vs whatever it was doing).
Updated docs in general,  added info on patching.

### DIFF
--- a/deployments/common/image/Dockerfile.trailer
+++ b/deployments/common/image/Dockerfile.trailer
@@ -28,10 +28,10 @@ RUN /opt/common-scripts/fix-certs
 COPY --chown=${NB_UID}:${NB_GID} environments/test    /opt/environments/test
 
 # Install security patches now that build is complete;  potentially this could be done sooner
-RUN apt-get update && \
-    apt-get dist-upgrade --yes && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+# -- RUN apt-get update && \
+# --     apt-get dist-upgrade --yes && \
+# --     apt-get clean && \
+# --     rm -rf /var/lib/apt/lists/*
 
 # For standalone operation outside JupyterHub,  note that  /etc also includes
 # common home directory files.
--- a/.github/workflows/image-build-test-scan.yml
+++ b/.github/workflows/image-build-test-scan.yml
@@ -1,13 +1,13 @@
 name:  Image Build, Test, Scan
 on:
-   schedule:
-     - cron: '0 0 */7 * *'
+   schedule:  # weekly
+     - cron: '0 0 * * 0'
    push:
-      # branches:
-      #   - main
+      branches:
+        - main
    pull_request:
-      # branches:
-      #   - main
+      branches:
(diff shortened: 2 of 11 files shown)


deployments/common/image/Dockerfile.trailer
```diff
@@ -28,10 +28,10 @@ RUN /opt/common-scripts/fix-certs
 COPY --chown=${NB_UID}:${NB_GID} environments/test    /opt/environments/test
 
 # Install security patches now that build is complete;  potentially this could be done sooner
-RUN apt-get update && \
-    apt-get dist-upgrade --yes && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+# -- RUN apt-get update && \
```
cslocum, line 31: I might be useful to add a comment here explaining why these lines are commented out but left in the file.  Ditto for all the mission-specific files.

cslocum:
Shouldn't the `apt-get dist-upgrade` removals be in Dockerfile.custom instead of Dockerfile, since Dockerfile gets regenerated each time the image is built?

cslocum:
A very minimal patch script example may be useful.  Do you have one you tested with?

cslocum:
This looks fantastic Todd! Nice work!

deployments/common/image/Dockerfile.trailer
jaytmiller: I left them there was as a reminder of what we used to do until our build process stabilizes,  including your scipy-notebook base image build.   I was thinking the base image build would make the above dist-upgrade unnneeded...  but wanted to see it in action first.

jaytmiller:
The _apt-get dist-upgrade_ was originally in the Dockerfile.trailer where it would occur after the build was basically complete.   When I commented it out in Dockerfile.trailer,  all the Dockerfiles were regenerated from a combination of Dockerfile.custom (unchanged) and Dockerfile.trailer.

jaytmiller:
Take a look at doc/SCRIPTS.md and the section on image patching.   In a nutshell,  I took the apt-get snippet out of Dockerfile.trailer and made _tools/apt-upgrade_ as an example of a patch script (which does something useful).

cslocum:
> The apt-get dist-upgrade was originally in the Dockerfile.trailer where it would occur after the build was basically complete. When I commented it out in Dockerfile.trailer, all the Dockerfiles were regenerated from a combination of Dockerfile.custom (unchanged) and Dockerfile.trailer.

That does make sense.  Thanks.  I think this is fine to merge.

PR #77, opened 2021-05-03
Only use 1 more IP address than is needed at all times

cslocum:


cslocum:
This is untested.  I will pull this into the next deployment I made to test it out.

jaytmiller (approved):
LGTM.    I think we should merge it and try a jwebbinar scaling test soon,  well before Thurs.

PR #80, opened 2021-05-07
Image patching additional scripts

jaytmiller:
Added image-delete script for deleting images in ECR using digests or tags.
Renamed to image-patch to patch-image.
Added shortcut scripts patch-os and patch-ssl for the most common patches.
Note that OS patches may break SSL and if so patch-ssl is intended to fix or update certs.


cslocum:
Fantastic.  I love these scripts.  Merge away!

jaytmiller:
Merging as suggested

PR #81, opened 2021-05-10
commit with new preloaded dirs in place. Requires PVC to be set in env

super-cob:


jaytmiller (commented):
Should these changes be someplace more generic than Roman DEV?
Is there a process for building the PVCs?

cslocum:
I agree with both of Todd's comments.   Can we move this code to the Roman common.yaml file please?  There are things in the Docker image that relate to/depend on these volumes, so each of the dev/test/prod clusters will need to have these configurations.  I'd like to see this done before merging if possible.

I'd also like to know more about the process.  Is there at least minimal documentation somewhere?

cslocum:
Can we terraform the infrastructure required for this to work?

super-cob:
I agree that it would be fantastic to automate some or all of this. At present the config is put in to envs where the PVC is expected to already exist and it would be getting ahead of yourself to put it in other envs because they would break. Instead, 1. decide whether to automate and how (terraform?) 2. push out the infrastructure half of this work to other envs. 3. NOW update configs, prioritizing factoring out this code into common.yaml (but I wouldn't go to all.yaml)

PR #83, opened 2021-05-17
Replace nbgitpuller

cslocum:
Replace nbgitpuller with a single `git-sync` script.

Also included is Todd's `efs-set-crds` script which initializes the CRDS persistent volume claim.


jaytmiller:
The efs-setup-crds is a WIP,  not even a draft.   But other than that LGTM.

jaytmiller (approved):
LGTM

PR #84, opened 2021-05-18
Add shared crds

jaytmiller:
Changes to support a preloaded crds cache symlinked into a user's $HOME/crds_cache from a shared readonly PVC.


cslocum (approved):
Looks good to me!

jaytmiller:
Everybody approved.

PR #86, opened 2021-05-25
Use email for auth/admin display

cslocum:
Also, set the default conda env to be jwebbinar.

Note that for other deployments, ENV.yaml in the secrets repo needs to have `username_key: eppn` removed (https://github.com/spacetelescope/jupyterhub-deploy/compare/use-email-for-admin?expand=1#diff-bbc8e85c2f3f01a0be886f3830a3bd65840d3dd8cafff5f87f42d915c52ef733L35)


jaytmiller (approved):
Can't really speak to the auth changes but the default environment setting is what I had in mind myself.  LGTM

PR #87, opened 2021-06-04
Tike image,  ds9,  lab setup,  ssl cert,  post-start-hook

jaytmiller:
Tike
----
Re-factored TIKE package dependencies to match updated notebooks from the
  spacetelescope/tess_content and spacetelescope/notebooks/MAST/TESS repo.
Added Tike environments/post-start-hook, replacing setup-notebooks,
  with the general charter of doing any initialization common to both
  JH and standalone images, e.g.  it also includes the rsync to $HOME.
  Also modified roman and tike common.yaml to call it's own verion of
  post-start-hook.  The overall post-start-hook scheme is needed to
  support standalone operation for CI tests on GitHub.
Added exoplanet kernel to Tike to accomodate Tensorflow
  incompatibilities (<2 vs. >2) with eleanor vs exoplanet.
Dropped astroconda from tess.yml channel setup.
Added common-scripts/env-src-install to install source tarballs from
  PyPi and used it to install dependencies for Tike's juliet package
  to ensure build against correct version of numpy api.

General / Common
----------------
Factor out jupyterlab extension setup into common-scripts/jupyter-lab-extension-setup.
Refactor ds9 setup to use Ubuntu's saods9 package vs. a conda env,  add pyds9
  to all environments via install-common.
Added image-develop script to support incremental development of
  mission requirements in a container launched from the common base image.
  This enables manually executing different environment setup steps to try out ideas
  rather than constantly rebuilding layers which don't really need it.
Modify fix-certs to replace ALL .pem files in /opt/conda with the
   STScI AWS cert.   This fixes image-test running on AWS CI-nodes.
Refactored cert setup so it can be repeated after Ubuntu updates which can undo cert setup.
Updated image-build with buildkit settings needed to avoid failures due to log sizes > 1 M.
Removed commented-out fix-permissions steps from Dockerfile.trailer.
Simplified image identifiers for standalone images:  drop unused account/registry name.
Defined CFLAGS in common/Dockerfile to handle gcc -fcommon flag needed
  for change -fno-common default which results in link time errors for
  old C-code which defined globals w/o using extern with duplicate
  global var definitions.
Update common astroquery.cfg removing JWST OPS MAST URL.  Add common .aws/config to define region=us-east-1.
Constrain sphinx<4.0.0 in common-env
Modify sscan-run-safety to handle 0 pip dependencies case
Fix weird @ file:/// style versions in pip freeze output in sscan-safety.
Modified common-scripts/test-notebooks to enable running a .py much
  like a .ipynb with the same handling of the source directory and
  outputs / temp directory.  Added --no-progress-bar in
  test-notebooks.
Added env-run to run a command in the specified environment.
Commented out Anchore image scan GitHub Action due to storage space failures.
Tweaks to image-build disk space recovery in GitHub Actions.
Increase GitHub swap to 24G due to roman Docker crash exit 137
Added safety and bandit to jupyterhub-deploy requirements.txt so they get installed on CI-nodes.
Added common .aws/config to default-home-contents since region us-east-1 is
  needed for some Tike test notebooks.
Fix git merge issue in LICENSE
Added pipdeptree to all environments to help sort out dependency issues.
Modified fix-certs to enable fixing conda certs whenever a new environment is
  created.  Factored out fix-conda-certs since it needs to execute as jovyan
  and not root as-is required by the system certs in /etc.   This solves the
  issue of lab extension install warnings due to certs.
Generalized jupyter extension pip package installation in common-env/jupyter.pip
  as well as jupyter-lab-extension-setup based on jdaviz.
Added post-start-hook call to Dockerfile.trailer so we get consistent behavior
  across all environments.  For jwebbinar post-start-hook is just a shim around
  setup-notebooks which should eventually be dropped once common.yaml is refactored.
Updated to jupyterlab-3.0.12

Jwebbinar
---------
Dropped jwebbinar kernel back to python-3.8 after compile issues.
Added mkdir to jwebbinar setup-notebooks test setup.
Modified jwebbinar Dockerfile to automatically find correct version of python for sed command.
Updated jwebbinar frozen specs for safety dependency scan.
Updated jwebbinar setup-notebooks tests so CI is testing the current set.
Added post-start-hook shim around setup-notebooks called from Dockerfile.trailer

Roman
-----
Convert roman to using post-start-hook.
Update roman for passing CI,  most/all notebook tests not yet defined.
Removed mirage kernel from roman image since it has no data and the
  image is frequently failing to "export" from Docker after building
  on GitHub, dying with exit code 137.


deployments/roman/image/Dockerfile
```diff
@@ -70,6 +46,62 @@ RUN cd /tmp && \
...
+
+# ----------------------------------------------------------------------
+# Run the hook script for updates which are common to both the standalone
+# image and to hub cloud operation.
+#
+USER root
+RUN chown -R ${NB_UID}:${NB_GID}  /home/${NB_USER}
+
+USER ${NB_UID}
+RUN /opt/environments/post-start-hook
```
jaytmiller, line 103: This really shouldn't be here anymore,  post-start-hook needs to be called from Dockerfile.trailer so now is.

deployments/common/image/common-scripts/fix-certs
```diff
@@ -1,10 +1,26 @@
-#! /bin/bash -eu
```
cslocum, line 1: Are failures OK here?

deployments/roman/image/Dockerfile
jaytmiller: Fortunately,  it's just the Dockerfile which needs to be regenerated,  Dockerfile.custom has already been updated.

deployments/common/image/common-scripts/global_bashrc
```diff
@@ -7,3 +7,5 @@ export PS1="${debian_chroot:+($debian_chroot)}\u@notebook:\w\$ "
 conda activate base
 
 export PATH="/opt/TurboVNC/bin:$PATH"
+
+alias ls="ls --color=auto"
```
cslocum, line 11: Thank you :)

deployments/common/image/default-home-contents/.astropy/config/astroquery.cfg
```diff
@@ -1,5 +1,216 @@
...
+
+# The Astrometry.net API key
+api_key =
+
 [mast]
 
 # Override MAST server with JWST A-string so we have access to
 # relevant data:
-server = https://pwjwdmsauiweb.stsci.edu
+# server = https://pwjwdmsauiweb.stsci.edu
```
cslocum, line 216: Why was this done?

deployments/common/image/common-scripts/fix-certs
jaytmiller: No.   I just moved the -eu down to line 7 on the right so that it executes after system RC files have already been sourced.  Could easily be moved back,  the main benefit of that switch is for scripts in tools... which when executed on OS-X cause bogus warnings/errors since system RC files with undefined variables are sourced after the shebang -eu is already in effect.  Moving -eu down OS-X stops bogus complaints so I changed this out of habit...  on Linux its really irrelevant I think.

deployments/jwebbinar/config/common.yaml
```diff
@@ -21,7 +21,7 @@ jupyterhub:
         c.KubeSpawner.lifecycle_hooks = {
             "postStart": {
                 "exec": {
-                    "command": ["/bin/sh", "-c", "rsync -r /etc/default-home-contents/ /home/jovyan; /opt/environments/setup-notebooks;  chmod +x /opt/conda/envs/jwebbinar/bin/ds9; rm -rf /home/jovyan/.local/share/jupyter/kernels/ds9; rm /home/jovyan/Desktop/DS9.desktop; chmod +x /home/jovyan/Desktop/*.desktop; rm -f /home/jovyan/.jdaviz-example-data; rm -f /home/jovyan/Desktop/build.out;  /opt/common-scripts/symlink-crds;"]
```
cslocum, line 24: The `rm`s are here because unfortunately, on the JWebbinar platform, existing users may already have these files in their environments.  Not sure how else to handle that problem, but we probably need a better solution as more platforms "go live".

deployments/common/image/default-home-contents/.astropy/config/astroquery.cfg
jaytmiller: Good question.   While updating astroquery.cfg,   I figured it was a 1.5 year old customization which was probably no longer valid.   But I should have researched more.   Fine to revert it,  your call.

deployments/jwebbinar/image/Dockerfile.custom
```diff
@@ -94,31 +85,12 @@ USER root
...
-
-COPY  --chown=${NB_UID}:${NB_GID} environments/setup-notebooks  /opt/environments
-RUN  /opt/environments/setup-notebooks
+# ----------------------------------------------------------------------
+COPY --chown=${NB_UID}:${NB_GID} environments/ /opt/environments
 
 # Fix JS encoding issue that prevents widgets from displaying properly in notebook mode
 # See https://github.com/jupyter/notebook/issues/6033 for details
-RUN sed -i 's/data-base-url="{{base_url | urlencode}}"/data-base-url="{{base_url}}"/g' ${CONDA_DIR}/lib/python3.9/site-packages/notebook/templates/notebook.html
+RUN sed -i 's/data-base-url="{{base_url | urlencode}}"/data-base-url="{{base_url}}"/g' `ls -1 ${CONDA_DIR}/lib/python*/site-packages/notebook/templates/notebook.html`
```
cslocum, line 93: Should this be in common now?

deployments/common/image/default-home-contents/.astropy/config/astroquery.cfg
gsnyder206: Please leave that server override commented out -- it breaks astroquery functionality in all the envs.  We may need to do more with custom MAST servers down the road with Roman, but for now this customization should be omitted.

deployments/jwebbinar/image/environments/post-start-hook
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
+
+set -eu
+
+# This is a placeholder version of post-start-hook which eventually should
+# fully replace setup-notebooks once common.yaml and this file are appropriately
+# refactored.   This file should container only those functions which need to
```
cslocum, line 7: container --> contain

deployments/jwebbinar/image/environments/setup-notebooks
```diff
@@ -13,6 +13,9 @@
 
 #/opt/common-scripts/set-notebook-kernel jwebbinar JWebbinar jwebbinar_prep/pipeline_products_session/*.ipynb
```
cslocum, line 15: Completely unrelated to this PR, but I've been thinking of requesting that we have text files in each jwebbinar_prep directory specifying which kernel should be used.

It's a pain to always be updating this.

deployments/jwebbinar/config/common.yaml
jaytmiller: One possible alternative is to wait until we're ready to field the new image,  then do a  "find /efs ..." on the CI-node and carefully rm them for all users.   Your call,  don't care either way.

deployments/jwebbinar/image/Dockerfile.custom
jaytmiller: Good catch.  I'll move it.

deployments/tike/image/Dockerfile
```diff
@@ -40,22 +20,69 @@ RUN curl --silent --show-error https://www.astro.princeton.edu/~jhartman/vartool
...
+RUN /opt/common-scripts/install-common tess
+
+COPY --chown=${NB_UID}:${NB_GID} environments/tess/basic.pip  /opt/environments/tess/basic.pip
+RUN /opt/common-scripts/env-update  tess /opt/environments/tess/basic.pip
+
+COPY --chown=${NB_UID}:${NB_GID} environments/tess/tess.pip  /opt/environments/tess/tess.pip
+RUN /opt/common-scripts/env-update  tess /opt/environments/tess/tess.pip
+
+# ----------
+# juliet equires batman-package and radvel but as supplied by PyPi they are currently built with
```
cslocum, line 46: "requires"

tools/image-dev
```diff
@@ -0,0 +1,51 @@
+#! /bin/bash
```
cslocum, line 1: This is fantastic!!!!

jaytmiller: Yeah,  give it a try sometime,  under the right conditions it's useful.

tools/patch-ssl
```diff
@@ -34,25 +34,15 @@ fi
...
 cat >patch-cert  <<EOF
-cat >tls-ca-bundle.pem <<END_CERT
+cd /etc/ssl/certs
+cat >stsci-tls-ca-bundle.pem <<END_CERT
 EOF
 
 cat ${CERT_FILE} >>patch-cert
 
 cat >>patch-cert  <<EOF
 END_CERT
```
cslocum, line 44: Can you help me understand what's going on with these nested EOFs please?

deployments/common/image/common-scripts/fix-certs
cslocum: Ah, I see.  That makes sense.

deployments/jwebbinar/config/common.yaml
cslocum: That's a great idea.

tools/patch-ssl
jaytmiller: Yeah.   A would-be-nice feature missing from the current patch system is the ability to include auxiliary files in addition to the patch script.   patch-ssl generates the patch-cert script which will be run inside the container from a little boilerplate and the new cert file.     Since there is no way to include aux files,  the cert is just inlined in the patch-cert script.    EOF only exists in patch-ssl.   END_CERT is written into patch-cert.   If you squint,  you can see there are 3 statements in patch-ssl which build up patch-cert something like:  boilerplate + new-cert + boilerplate.

deployments/jwebbinar/image/Dockerfile.custom
jaytmiller: Moved.

tools/patch-ssl
cslocum: >  EOF only exists in patch-ssl. END_CERT is written into patch-cert

OH, ok.  Thanks :)  That was hurting my brain.

cslocum (approved):
This is fantastic!  No merge objections.

PR #88, opened 2021-06-09
Update notebook config for JWebbinar-4; Fix Lab extensions for JupyterLab==3.0.12

cslocum:


deployments/common/image/base.yml
```diff
@@ -4,8 +4,8 @@ channels:
   - http://ssb.stsci.edu/astroconda
   - defaults
 dependencies:
-  - bqplot-image-gl
-  - bqplot
+  #- bqplot-image-gl==1.4.1
```
jaytmiller, line 7: We should consider reconciling these with common-scripts/jupyter-lab-extension-setup and common-env/jupyter.pip.   Some packages might move,  others may already be there and would just be deleted here.  The common Dockerfile does install-common on the base env so standard lab setup applies.

deployments/jwebbinar/image/environments/setup-notebooks
```diff
@@ -1,6 +1,4 @@
-#! /bin/bash -eu
-
-# This is a standard entry-point used to set up notebooks for this
+s is a standard entry-point used to set up notebooks for this
```
jaytmiller, line 1: You're probably already aware but some of the notebooks defined as tests below didn't pass,  including one failure due to a missing package (stcal) which might be easy to fix.

jaytmiller (approved):
A couple comments to consider but LGTM,  merge as-is if this is the dry-run code.

deployments/common/image/base.yml
cslocum: Agreed.  I think for now though I'm not going to mess with things for webbinar-4.  In case I need to rebuild, I'd rather keep things as close to same as possible.

cslocum:
Thanks Todd - I am going to merge as is.  This reflects what's being used for JWebbinar-4.  But I'll make the corrections pointed out above for the next one.

PR #89, opened 2021-06-10
Image squash R&D prototype

jaytmiller:
Added crosslink program for hard linking duplicate conda files.
Added draft image squash scheme based on hard linking /opt/conda and then
  squashing the resulting image using docker export/import.  This is
  pre-alpha code but seems to work fairly well.

Not positive we should merge this but thought we should have the option.


cslocum (approved):
This all looks really great!  Bent by brain a bit reviewing it :)  Nice work.

I think this can be merged.  You have only introduced scripts here, so there's no harm in having them in main (your PR comment shows hesitance).

Have you see any ill effects of applying these?

jaytmiller:
I haven't seen any negative effects,  it's mainly a matter of clutter.   Nothing happens unless you type "squash".   I should probably write a little section in SCRIPTS.md.  If I don't touch anything else it shouldn't cause conflicts.

PR #91, opened 2021-06-14
Very minor cleaup after aborted attempt at VNC copy/paste

cslocum:
Really not much to say...just want to get the latest into main.


deployments/roman/image/environments/frozen/mirage-frozen.yml
```diff
@@ -0,0 +1,366 @@
+name: mirage
```
jaytmiller, line 1: Because Roman was having trouble building on GitHub,  and because it was just a data-less paperweight,   I dropped the Mirage kernel from Roman.   So in theory this should not be here any more.   IDK if it will cause trouble or not.

deployments/common/image/environments/frozen/ds9-frozen.yml
```diff
@@ -0,0 +1,151 @@
+name: ds9
```
jaytmiller, line 1: The DS-9 kernel is now gone and replaced by the Ubuntu package for DS-9.   This should not be here.

jaytmiller (approved):
Drop the two no-longer-valid/used frozen specs and LGTM.

cslocum:
Thanks Todd.

> Drop the two no-longer-valid/used frozen specs and LGTM.

Done.

PR #95, opened 2021-06-18
Hook up central ECR

cslocum:
Light refactor to allow developers to choose to use either a central ECR repository or the repository local to the account.

Add a new script for image promotion: `image-promote`
Add new script, mainly for testing, that removes a tag from an image: `image-remove-tag`


deployments/roman/image/environments/post-start-hook
```diff
@@ -25,7 +25,8 @@ rsync -r  /etc/default-home-contents/ /home/jovyan
 
 # .............................................................................................
 
-git clone https://github.com/spacetelescope/romancal-notebooks.git
+# this will fail if the clone already exists
+git clone https://github.com/spacetelescope/romancal-notebooks.git || true
```
jaytmiller, line 29: Is there a reason not to use git-sync for this?  git clone was an oversight.

tools/image-remove-tag
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
```
jaytmiller, line 1: For this I don't think we need both the image sha1sum and tag.  The same tag cannot exist twice.  I may be missing something.

Other than needing to correctly redefine IMAGE_REPO,  image-delete is more in line with what untagging does unless there are multiple tags.   In theory it also loops over a list  of tags or sha1sums,  and works on either without defining both.

In any case,  we should probably just do one or the other,  but not a big deal which.

jaytmiller (approved):
LGTM.   Consider building in awsudo unless there's a reason not to.   Consider using image-delete instead of image-remove-tag.

deployments/roman/image/environments/post-start-hook
cslocum: Ah, OK.  I can certainly do that.

tools/image-remove-tag
cslocum: Do you suggest updating `image-delete` to use IMAGE_REPO?  I'm fine with that, and ditching this script.

doc/SCRIPTS.md
```diff
@@ -25,12 +25,14 @@ Automated testing:
...
 
-Interact with ECR to push images, pull images, and delete images:
+Interact with ECR to push images, pull images, delete images, and tag/untag images:
 
 - image-push    -- push the built image to ECR at the configured tag
 - image-pull    -- pull the configured image tag from ECR to the local Docker
 - image-delete  -- delete the specified image tags or digests from ECR, e.g. to ditch vulnerable images
 - image-login   -- log in to ECR
+- image-promote -- promote an image to the next tier (e.g. dev --> test); use with "awsudo $ADMIN_ARN"
+- image-remove-tag -- remove a tag from a specified image; use with "awsudo $ADMIN_ARN"
```
cslocum, line 35: (Note to self: remove this line)

tools/image-remove-tag
jaytmiller: Yeah,  whatever tweaks make it as easy as possible to use tags,  and just the tag:  (something)-latest or (something)-unscanned) or whatever.  Something simpler than $IMAGE_ID.

cslocum:
> Consider building in awsudo unless there's a reason not to

The reason for this is simply that I'm using boto3, not executing shell commands.  I do prefer to use boto3, but if you think it's a big enough deal, I can certainly re-write it without boto3.

cslocum:
Going to merge, even with `awsudo` still being required for `image-promote`.  I filed https://jira.stsci.edu/browse/JUSI-716 to address it in the future.

PR #96, opened 2021-06-18
Added apt-install script to simplify/standardize apt-installs

jaytmiller:
Added apt-install script to simplify apt-get installs with standard
  setup and cleanup based on a simple package list.
Refactored common image apt packages into multiple apt-install's
  revolving around functional topics,  particularly to be able
  to see topic space consumption in docker history.


cslocum (approved):
Awesome.  Good stuff here.  No merge objections.

PR #102, opened 2021-07-02
Fix git-sync

cslocum:
`git-sync` was creating backups of files that had no changes other than whitespace (between the repo and the clone in the image).

Adding `-w` to `git diff` in the script fixes this problem.


jaytmiller (approved):
Looks great,  awesome to figure this one out!

PR #105, opened 2021-07-12
git-sync ignores whitespace diffs.

cslocum:
We were having a problem where diffs were being detected (technically, properly), but the only differences were whitespace.  This was resulting in `git-sync` creating backups of the "changed" files - an unwanted behavior.  Changes in this PR adjust some logic, and add a flag to `git diff` to ignore whitespace.

Also fixed logic error that evaluates `$USE_CENTRAL_ECR`


deployments/common/image/base.yml
```diff
@@ -4,8 +4,8 @@ channels:
   - http://ssb.stsci.edu/astroconda
   - defaults
 dependencies:
-  #- bqplot-image-gl==1.4.1
```
cslocum, line 7: @jaytmiller I'm not sure why these packages needed to be specified here, but they were necessary in the latest jwebbinar build.

jaytmiller: I think packages in base.yml are only going into the base Python environment.

Aside from env cloning,  packages to be installed in all Python environments are either defined in common-env or common-scripts.

jaytmiller (commented):
Check into difference between base.yml and packages spec'ed in common-env or common-scripts.

PR #107, opened 2021-07-13
Script for deleting ECR images that are failing security scans

cslocum:
Directly from the docstring:

> This tool scans an ECR repository to find images with vulnerabilities.
> Images that are tagged are ignored.  All untagged images that are found to have
> vulnerabilities are considered to be candidates for deletion.
>
> Interactive and automated actions are both supported.  By default, this tool
> will ask you to confirm deletion of each image that has been found to be
> insecure.
>
> Run "clean-ecr --help" for usage information.

I tested both interactive and automated modes on the local Roman DEV account (not the central ECR account).

I also changed some of the global variable names to make more sense.  I've grep'd to see if these changes will impact any other scripts, and I don't think they will.

NOTE: I took the approach of ignoring all tagged images.  We will be notified via email if a tagged image is failing security scans, and that would require manual intervention.

### DIFF
--- a/derived-env
+++ b/derived-env
@@ -4,16 +4,18 @@
 # automatically sourced into setup-env
 
 # Octarine AWS JupyterHub setup
-export ACCOUNT=${ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
-export ECR_ACCOUNT=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
+#export ACCOUNT=${ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
+#export ECR_ACCOUNT=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
 
 if [ "$USE_CENTRAL_ECR" = true ]
 then
-    export ECR_ACCOUNT_TO_USE=$ECR_ACCOUNT
+    export ECR_REGISTRY=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
     export IMAGE_REPO=${DEPLOYMENT_NAME}
+    export ECR_ACCOUNT_TO_USE=${CENTRAL_ECR_ACCOUNT_ID}
 else
-    export ECR_ACCOUNT_TO_USE=$ACCOUNT
+    export ECR_REGISTRY=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
     export IMAGE_REPO=${DEPLOYMENT_NAME}-user-image
+    export ECR_ACCOUNT_TO_USE=${ACCOUNT_ID}
 fi
 
 export JUPYTERHUB_DIR=`pwd`
--- a/derived-env
+++ b/derived-env
@@ -7,13 +7,13 @@
 #export ACCOUNT=${ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
 #export ECR_ACCOUNT=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
 
-if [ "$USE_CENTRAL_ECR" = true ]
+if [ "$USE_CENTRAL_ECR" == "true" ]
 then
     export ECR_REGISTRY=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
     export IMAGE_REPO=${DEPLOYMENT_NAME}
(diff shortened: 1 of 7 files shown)


tools/clean-ecr
```diff
@@ -0,0 +1,115 @@
...
+    valid_inputs = {'y': True,'n': False}
+    prompt = f'Delete vulnerable image with digest {img.digest}? [y/n]'
+
+    while True:
+        sys.stdout.write(prompt)
+        response = input().lower()
+        if response in valid_inputs:
+            img.to_delete = valid_inputs[response]
+            break
+        else:
```
cslocum, line 64: Perhaps this "else" block is not needed?

derived-env
```diff
@@ -4,16 +4,18 @@
...
 # Octarine AWS JupyterHub setup
-export ACCOUNT=${ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
-export ECR_ACCOUNT=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
+#export ACCOUNT=${ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
+#export ECR_ACCOUNT=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
 
 if [ "$USE_CENTRAL_ECR" = true ]
 then
-    export ECR_ACCOUNT_TO_USE=$ECR_ACCOUNT
+    export ECR_REGISTRY=${CENTRAL_ECR_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com
```
jaytmiller, line 12: Is ECR_REGISTRY used anywhere or is it a loose end?

tools/clean-ecr
```diff
@@ -0,0 +1,115 @@
...
+vulnerabilities are considered to be candidates for deletion.
+
+Interactive and automated actions are both supported.  By default, this tool
+will ask you to confirm deletion of each image that has been found to be
+insecure.
+
+Run "clean-ecr --help" for usage information.
+"""
+
+admin_arn = os.environ.get('ADMIN_ARN')
```
jaytmiller, line 21: Use upper case for globals, always,  not as I sometimes don't do.

tools/clean-ecr
```diff
@@ -0,0 +1,115 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Find and delete images in ECR that are failing security scans and are untagged'
+    )
+    parser.add_argument(
+        '--all', dest='delete_all', action='store_true', default=False,
+        help='don\'t prompt whether to delete each image, just delete all untagged insecure images'
+    )
+    parser.add_argument(
```
jaytmiller, line 86: Get "Focal" from env IMAGE_UBUNTU_NAME.  It's "Groovy" now.  Might be other code to fix like image-scan-report.

tools/clean-ecr
```diff
@@ -0,0 +1,115 @@
...
+        help='Ubuntu version substring (default is "Focal")'
+    )
+    parser.add_argument(
+        '--min-severity-level', dest='min_severity', action='store', default='HIGH',
+        help='minimum severity level of vulnerabilities (default is "HIGH")'
+    )
+    args = parser.parse_args()
+
+    # find candidates for deletion (all untagged images in a repo)
+    for d in get_image_digests():
```
jaytmiller, line 97: I'd put the code from here down into main() and pass it all, os_version, and min_severity_level.

jaytmiller (approved):
Minor comments,  seems great to me.

derived-env
cslocum: I will double check...

tools/clean-ecr
cslocum: Ah!  Thanks, will do.

tools/clean-ecr
cslocum: Good idea.

derived-env
cslocum: Yes, it's used in a couple of `image-*` scripts.

tools/clean-ecr
```diff
@@ -0,0 +1,150 @@
...
+            scan_output, err = proc.communicate()
+            if proc.returncode == 0:
+                self.insecure = False
+        except Exception as e:
+            raise e
+
+        if self.insecure:
+            print()
+            # THIS IS EMPTY
+            print(scan_output)
```
cslocum, line 62: @jaytmiller @jmatuskey - can either of you spot why this output may be empty?  `image-scan-report` should always produce an output.

[I had realized that I was flagging everything as "deletable" if the output was non-0, but I was checking against the specified error level, so I am trying to actually read in the output to parse through it.]

PR #108, opened 2021-07-16
Update clean-ecr to not fetch/consider Ubuntu findings by default

cslocum:
This also makes the default mode much faster.


jaytmiller (commented):
Looks great to me!

PR #112, opened 2021-07-20
Version tools

jaytmiller:
Added some scripts used to dump s/w versions:

1. Inside the image
2. On the host

Scripts are:

1. tools/versions  (combined image + host)
2. tools/versions-image
3. tools/versions-host
4. /opt/common-scripts/dump-version  (runs inside notebook container)

Tweaked handling of IMAGE_TAG so it is set explicitly in setup-env to: 
    e.g. unscanned-latest-dev
and used unmodified for scripts which refer to it, IMAGE_ID, etc.

Note that config YAML files still refer to the approved image tags: 
   e.g.  latest-dev

### DIFF
--- a/tools/versions-host
+++ b/tools/versions-host
@@ -100,6 +100,7 @@ def main():
     host = dict(
         jupyterhub_deploy = get_git_version("/home/ec2-user/jupyterhub-deploy"),
         terraform_deploy = get_git_version("/home/ec2-user/terraform-deploy"),
+        jupyter_docker_stacks = get_git_version("/home/ec2-user/docker-stacks"),
         terraform = get_terraform(),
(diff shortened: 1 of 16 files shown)


tools/versions-host
```diff
@@ -0,0 +1,113 @@
...
+
+from collections import defaultdict
+from contextlib import contextmanager
+
+import yaml
+
+
+
+for envvar in ["ADMIN_ARN", "DEPLOYMENT_NAME"]:
+    globals()[envvar] =  os.environ[envvar]
```
cslocum, line 20: Nice!  This is much cleaner than what I usually do :)

cslocum (approved):
Very nice!  This all looks great and useful.

PR #113, opened 2021-07-26
Image promotion no longer requires awsudo

super-cob:


cslocum (approved):
This looks great!  Thanks.

PR #114, opened 2021-07-27
Deactivate failing Tike test notebook priority 1

jaytmiller:
Without the installation of eleanor or other similar changes,  the priority 1 notebook for Tike is not going to pass.   Deactivate for now.

### DIFF
--- a/tools/image-build
+++ b/tools/image-build
@@ -19,7 +19,7 @@
 # as a solution to build fail like:
 # failed to export image: failed to create image: failed to get layer sha256:...: layer does not exist
 
-# LOG size settings are a solution to build failuyre with this error:
+# LOG size settings are a solution to build failure with this error:
 #   [output clipped, log limit 1MiB reached]
 
 export DOCKER_BUILDKIT=1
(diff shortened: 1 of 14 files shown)


tools/image-build
```diff
@@ -33,22 +33,27 @@ image-update
...
+image-base
 
 
 # ..................................................................................
 # Build STScI shared layer for all deployments
 cd ${COMMON_DIR}
 echo "========================= Building Common Image ========================"
-time docker build --tag ${COMMON_ID} --build-arg USE_FROZEN=${USE_FROZEN} .
-
+time docker build --progress plain --tag ${COMMON_ID} \
```
cslocum, line 43: Does adding `--progress` slow down the build at all?  Probably not, just wondering.

cslocum (approved):
This looks good to me!  Merge away.  Did leave one question.

tools/image-build
jaytmiller: Not that I noticed.

jaytmiller: We should be able to verify the timing once we get working builds again.

PR #116, opened 2021-08-04
Tike test ops setup

jaytmiller:
Changes made for basic Tike TEST and OPS account setup.
Added custom ui templates for Tike TEST and OPS which are required for deploy-all.
Added convenience aliases to make some terraform-deploy tasks easier.
Modified image-promote-testops to use tag latest-prod instead of latest,  consistent with jwebbinar and Tike.
Tweaked templates which used YAML list notation as placeholders making it impossible to round-trip through sops since the placeholder was pretty printed from [ REDACTED ] to - REDACTED on a new line which is no longer a valid format.
Currently fixes image name bug in deploy script where ECR_ACCOUNT_TO_USE needs to be replaced by ECR_REGISTRY,  probably covered by other concurrent PR's,  may disappear from this merge.

### DIFF
--- a/derived-env
+++ b/derived-env
@@ -42,3 +42,11 @@ export PATH="`pwd`/tools":${PATH}
 export TMOUT=7200
 export TF_LOG=TRACE
 export TF_LOG_PATH=terraform.log
+
+alias awsu="awsudo -d 3600 $ADMIN_ARN"
+alias terraform-init="awsu terraform init -backend-config=./backend.conf -backend-config=./backend.conf"
+alias terraform-apply="awsu terraform apply --var-file=${DEPLOYMENT_NAME}.tfvars"
+alias terraform-destroy="awsu terraform destroy -var-file=${DEPLOYMENT_NAME}.tfvars"
+alias helm-destroy="awsu helm uninstall ${DEPLOYMENT_NAME}-${ENVIRONMENT}"
+alias where="echo ${DEPLOYMENT_NAME}-${ENVIRONMENT}"
+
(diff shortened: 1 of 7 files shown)


derived-env
```diff
@@ -42,3 +42,11 @@ export PATH="`pwd`/tools":${PATH}
...
 export TF_LOG=TRACE
 export TF_LOG_PATH=terraform.log
+
+alias awsu="awsudo -d 3600 $ADMIN_ARN"
+alias terraform-init="awsu terraform init -backend-config=./backend.conf -backend-config=./backend.conf"
+alias terraform-apply="awsu terraform apply --var-file=${DEPLOYMENT_NAME}.tfvars"
+alias terraform-destroy="awsu terraform destroy -var-file=${DEPLOYMENT_NAME}.tfvars"
+alias helm-destroy="awsu helm uninstall ${DEPLOYMENT_NAME}-${ENVIRONMENT}"
+alias where="echo ${DEPLOYMENT_NAME}-${ENVIRONMENT}"
+
```
cslocum, line 52: Doesn't need to be done now, but probably we should update DEPLOYMENT.md to use these aliases.

derived-env
```diff
@@ -42,3 +42,11 @@ export PATH="`pwd`/tools":${PATH}
 export TMOUT=7200
 export TF_LOG=TRACE
 export TF_LOG_PATH=terraform.log
+
+alias awsu="awsudo -d 3600 $ADMIN_ARN"
```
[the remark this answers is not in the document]
jaytmiller: LOL

derived-env
jaytmiller: OK

cslocum (approved):
LGTM

PR #118, opened 2021-08-05
Deploy DataDog without hardcoding API key in Helm chart

cslocum:
~- Pull down the latest DataDog image from the public Amazon registry~
~- Build our own image on top of that, installing a tiny script that fetches the DataDog API key from AWS Secrets Manager and uses that key to push metrics to DataDog~
- Add a helm chart for deployment
- Add tool that grabs DataDog secret; this is used in the main deploy script
- Add a tool that deploys the DataDog to the cluster
- Very minimal, incomplete README that reflects reality :p

---

TODO:
~- [x] Update build/deploy script to also push image to central ECR repo; use `image-login`~
~- [x] Pull out secret ID from `get-api-key` and instead set an env variable that it can use; the env variable can probably been passed in from the CI node, so none of it needs to be in github - probably just EOF into a temp file before copy~
- [x] Cleanup/delete unneeded code
~- [x] Get new ECR registry; submitted ticket RITM0078629~
- [x] Test!  Verify that all metrics show up on DataDog dashboards
~- [ ] Ticket out-of-scope stuff like writing a tool to pull images down and deploy on other SDLC accounts, migrate DataDog stuff to dedicated repo --> update jupyterhub tooling to account for that, etc.~


cslocum:
Very much a work in progress still.  DataDog does deploy, but does not yet work since the image isn't in any repo.  But I wanted to PR this so folks can give feedback early...

jaytmiller (approved):
Looks really nice!

setup-env.template
```diff
@@ -12,6 +12,7 @@ export USE_CENTRAL_ECR=false
 export ADMIN_ROLENAME=aws-admin-role
 export DEPLOYMENT_NAME=cluster-name
 export ENVIRONMENT=sandbox  # sandbox, dev, test, or prod
+export DD_API_KEY_ID=look-in-aws-ssm
```
super-cob, line 15: since usually you export DATADOG_API_KEY and DATADOG_APP_KEY maybe we should keep it the same even though it's a long 4 word variable

cslocum: I don't think we're actually even setting an app key.  At least Fox hasn't been doing so.

cslocum: For reference, this is what he sent me: https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=DMD&title=DataDog+setup.

We can definitely add more later, but the criteria to be met here is simply to replace his hardcoded API Key with one that is fetched on-the-fly with proper credentials.

tools/deploy-datadog
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash -e
+
+cd ${JUPYTERHUB_DIR}/datadog
```
super-cob, line 3: this line isn't working for me while running `tools/deploy-datadog` from jupyterhub-deploy/

jmatuskey:
I just meant that the env. vars for datadog have DATADOG fully spelled out
so as a small thing maybe keep that naming convention the same

setup-env.template
super-cob: I meant name it DATADOG_API_KEY_ID

jaytmiller:
The latest version seems super clean,  nice PR!

PR #119, opened 2021-08-10
Disable loading /home/shared for Roman; update deploy-jupyterhub

cslocum:
`deploy-jupyterhub` needed to be updated to specify `jupyterhub.singleuser.image.tag` in the `helm install`.  Not quite sure why, but the image puller stopped using the specified singleuser image specified in our YAML config.

Also commented out all the extra volume stuff for preloaded files (for Roman).  We only have a volume for JWebbinar.


jaytmiller (approved):
LGTM.

PR #120, opened 2021-08-11
Add license to README

cslocum:


jaytmiller (changes requested):
It's a pain,   but I'd recommend adding the file LICENSE you can find here:  https://github.com/spacetelescope/jwst/blob/master/LICENSE.   I believe it's very similar to BSD 3 clause and we may be required to use it technically.

cslocum:
> It's a pain, but I'd recommend adding the file LICENSE you can find here: https://github.com/spacetelescope/jwst/blob/master/LICENSE. I believe it's very similar to BSD 3 clause and we may be required to use it technically.

Done, Merging.

PR #124, opened 2021-08-23
Fix image-scan already re-tagged

jaytmiller:
Modified image-scan-report to handle the case where an image has been re-tagged by the time an attempt to fetch a report is made so the default image tag is no longer valid.


cslocum (approved):
Looks great!

PR #127, opened 2021-08-27
Fix sscan-bandit missing some files

jaytmiller:
sscan-find-python files was missing scripts using _#!/usr/bin/env python_ (no space after _#!_)
added -v to sscan-bandit to get more feedback on scanning.


cslocum (approved):
Looks good.

PR #128, opened 2021-08-30
Tweaks related to simplify-env changes,  different env vars

jaytmiller:
Tweaks for image-release, image-promote due to setup-env changes.
Switched default tag from unscanned-latest-xxx to latest-xxx since only scanned images should be released.


cslocum (approved):
Looks good to me!

PR #130, opened 2021-09-02
Secure git-sync

jaytmiller:
Updated git-sync to address bandit security concerns,  subprocess calls using os.system() or shell=True.


cslocum (approved):
This looks good!

PR #131, opened 2021-09-02
Fix pip requirements freezing

jaytmiller:
Re-arranged storage of frozen requirements for all missions, build scripts.

So they're not wiped out by the final copy of all of environments to
  /opt/environments.  Frozen requirements are now stored in env-frozen and
  /opt/env-frozen.
Updates to all Dockerfile.custom and frozen requirements files using new scheme.
Drop .x file removal from image-build,  it screws up image-build-all.
Deleted obsolete env-create and env-clone scripts.
Update frozen requirements.
Overhaul image-cp,  was failing on GitHub.


deployments/common/image/Dockerfile.trailer
```diff
@@ -15,7 +15,7 @@ RUN /opt/common-scripts/kernel-setup   # set up Ipython / JupyterLab kernels
 
 USER root
 
-RUN chown -R ${NB_UID}:${NB_GID}  /home/${NB_USER}
+RUN find /home/${NB_USER} ! -uid ${NB_UID} | xargs chown ${NB_UID}:${NB_GID}
```
jaytmiller, line 18: This is for performance,  40 secs --> 2 secs

PR #139, opened 2021-09-09
Fix requirements freezing,  base image builds,  setuptools issue,  misc

jaytmiller:
Re-arranged storage of frozen requirements for all missions, build scripts.
Frozen requirements are now stored in env-frozen and /opt/env-frozen to facilitate
  copying updated versions out of the fully built image.  Were being overwritten
  by final blanket copy of environments.

Re-worked image-base due to changes in docker-stack Makefile,   removal of DARGS.

Pinned setuptools<58 due to removal of use_2to3 parameter required by many packages.
Refactored setuptools installation for better repeatability and to pin to ~= 57.0 in only one place.

Optimized chown /home/jovyan to jovyan,  40 sec -> 2 sec.
Drop .x file removal from image-build,  it screws up image-build-all.
Deleted obsolete env-create and env-clone scripts.
Unpinned Tike ML requirements due to vulnerabilities in Tensorflow and Keras reported by safety and dependabot.
Simplified / updated image-freeze and image-cp scripts.
Added text translation to "where".
Removed global_bashrc from jwebbinar environment.
Turned on plain logging in image-base


cslocum:
> Frozen requirements are now stored in env-frozen and /opt/env-frozen to facilitate
copying updated versions out of the fully built image. Were being overwritten
by final blanket copy of environments.

Ah!  That makes too much sense...

deployments/common/image/Dockerfile.trailer
```diff
@@ -15,7 +15,7 @@ RUN /opt/common-scripts/kernel-setup   # set up Ipython / JupyterLab kernels
 
 USER root
 
-RUN chown -R ${NB_UID}:${NB_GID}  /home/${NB_USER}
+RUN find /home/${NB_USER} ! -uid ${NB_UID} | xargs chown ${NB_UID}:${NB_GID}
```
cslocum, line 18: Nice.

deployments/common/image/common-scripts/env-compile
```diff
@@ -59,7 +60,11 @@ if [[ "$USE_FROZEN" != "1" ]]; then
     for pip in ${all_constraints}; do
         echo ${pip}   # show one per line.
     done
-    rm -f /opt/environments/${env}/requirements.txt
-    pip-compile  -v  --output-file /opt/environments/${env}/requirements.txt  --pip-args "${PIP_SWITCHES}"  ${all_constraints}
+
+    # If pip-compile finds an existing requirements.txt file that fulfils the
+    # dependencies then no changes will be made, even if updates are available.
```
cslocum, line 65: I'm confused about this.  If USER_FROZEN=0, then don't we want to get the updates?

jaytmiller: Negative logic was a poor choice there but it works.   The idea is that if USE_FROZEN=0,  (USE_FROZEN != 1) we're determining fully pinned requirements and should re-compile to update.   Otherwise the frozen requirements remain unchanged.  In either case,  the pip-sync command later installs what you ask for,  either newly updated or frozen and unchanged.

jaytmiller: The pip-compile related lines changed (look deleted) as well as moved down due to the addition of ${env_reqs}.

cslocum: Ah, I see, I didn't read it closely enough.  Thanks.

cslocum:
Looks great!  Nice work.  Merge away...

cslocum (approved):

PR #140, opened 2021-09-09
Update base.yml to get jupyterlab > 3.1.4,  update frozen requirements.

jaytmiller:
The key change is base.yml,  and the absence of frozen requirements which show the concrete effect of the change points out lack of solid pinning of the base image.  Checking manually,  the base environment now gets jupyterlab==3.1.11 like the mission environments.


cslocum (approved):
Looks good :+1:

PR #145, opened 2021-09-10
Improve safety output

jaytmiller:
Add explicit flushing of stderr to avoid confusing "ignored IDs" with "safety results" in output.


cslocum (approved):
Looks good!

PR #147, opened 2021-09-24
Jupyterlab extension and base environment setup consolidation.

jaytmiller:
Refactored common image some, distributing non-redundant conda base.yml into
  common-env package lists which are installed everywhere.  base environment is
  now updated using install-common which also performs central jupyter
  labextension, notebook, and npm setup once. install-common is now only used
  for base since more generally common-env/*.conda and common-env/*.pip are
  installed by env-conda and env-compile respectively.  jupyterlab extension
  setup is not performed for mission environments under the assumption that it
  is global.   documented rationales for specific package common-env jupyter
  package setup in common-env/PINNING.md.

Added code to freeze the base environment using conda env export into
  env-frozen/base/requirements.yml for each mission capturing both conda and
  pip setup for base.  When USE_FROZEN=1 base is updated by requirements.yml to
  pin versions and add missing packages after it has been built by
  scipy-notebook; this results in image bloat effectively duplicating base but
  should back-pin base environment conda and pip package versions.  base
  requirements are recorded for each mission seperately and the base environment
  is updated as the first action of each Dockerfile.custom.

Added frozen base env requirements for all missions,  updated requirements for
  all missions.

Added tensorflow 2.6.0 to Tike safety.ignore,  no effective work around yet.

Pinned git+https:// package references to pandeia-coronagraphy with @git-hash

Removed docs on original install framework in FRAMEWORK.md in favor of
  pip-tools only.

Removed comments using obsolete install framework from Roman Dockerfiles

Hacked libcurl in base env frozen requirements spec, unpinning, version frozen
  by conda env export is not available at time of frozen env update.  Looks like
  the .0 version was deleted in favor of .1...  but .0 still winds up in specs.


deployments/common/image/common-env/PINNING.md
```diff
@@ -0,0 +1,305 @@
...
+configure classic Jupyter Notebook and JupyterLab 3.0 to display
+ipywidgets. With pip, do:
+
+  `pip install ipywidgets`
+
+
+nbwidgetsextension
+------------------
+
+Is related ipywidgets.
```
cslocum, line 108: Missing "to"

deployments/common/image/common-env/PINNING.md
```diff
@@ -0,0 +1,305 @@
...
+jupyter-golden-layout
+---------------------
+
+bqplot-image-gl
+---------------
+   [bqplot-image-gl npmjs](https://www.npmjs.com/package/bqplot-image-gl)
+
+```
+   npm install --save jupyter-golden-layout
+```
```
cslocum, line 305: Fantastic doc!

deployments/common/image/common-env/common.pip
```diff
@@ -7,6 +7,7 @@ pyds9
 pipdeptree
 pip-tools
 pydeps
-graphviz
```
cslocum, line 10: Is this not needed?

jaytmiller: That's the coming change I mentioned.  It's needed by tools/image-graph-env to do dependency plots.   But there are issues installing it with conda vs. pip since it is named differently in each.  So my planned approach is to install it on the fly whenever someone wants to do a plot.   Still building and testing that.

deployments/jwebbinar/image/env-frozen/jwebbinar/requirements.txt
```diff
@@ -1205,6 +1212,7 @@ wheel==0.37.0
     # via pip-tools
 widgetsnbextension==3.5.1
     # via
+    #   -r /opt/common-env/jupyter.pip
     #   -r /opt/environments/jwebbinar/jdaviz.pip
```
cslocum, line 1216: This can be gotten rid of, the jdaviz kernel/env don't exist for jwebbinar anymore.  Pretty sure that was merged?

cslocum (approved):
This is really nice work!

deployments/common/image/common-env/common.pip
cslocum: Got it, thanks.

deployments/jwebbinar/image/env-frozen/jwebbinar/requirements.txt
jaytmiller: Not sure what you mean.  Those lines are in the pip freeze/compile output for jwebbinar.   They indicate that widgetsnbextension is being pulled in by both jupyter.pip and jdaviz.pip.   jdaviz.pip is presumably what is adding jdaviz to the jwebbinar environment.   If it is listed explicitly in jdaviz.pip,  it should probably be removed from that since it is now in common-env/jupyter.pip.

PR #148, opened 2021-09-24
Fix image-graph-env

jaytmiller:
Restored install of graphviz needed for image-graph-env dependency plotting
  by adding it dynamically for each plot.   It was removed due to package
  naming inconsistencies between conda and PyPi.


cslocum (approved):
Nice!

PR #166, opened 2021-11-05
Variable branch git-sync support

jaytmiller:
Added JH_CONTENT_VERSION env var used to define SDLC-specific values
  for content repo tags.
Defaults JH_CONTENT_VERSION=main in common Dockerfile.
JupyterHub YAML config like tike/config/prod.yaml overrides env var
  at runtime to something like jupyterhub-tike-ops.
Calls to git-sync in the container sync relative to $JH_CONTENT_VERSION
  if desired or continue to hard-code branch as they used to.


jaytmiller:
This should not be merged pending discussion of SDLC-account-specific branches.

cslocum:
I like this convention.  I know it needs more discussion, but I'm all on board with the SDLC-account-specific branches.

PR #167, opened 2021-11-05
Pinned Tensorflow and relatives ~= 2.7.x

jaytmiller:


jaytmiller:
I dropped Tensorflow for Tike back to 2.6 for that family of packages after discovering that Tensorflow~=2.7.0 broke at least one notebook.   Prior to this update,  Tike has a mismatched set of Tensorflow packages with some set to 2.7 and some set to 2.6.    This sets Tike everything uniformly to Tensorflow 2.6.

cslocum (approved):
Looks good to me 👍

PR #168, opened 2021-11-08
Put strings around awsudo calls (Required by awsudo v2)

cslocum:


jaytmiller (approved):
Looks great,  thanks!

PR #174, opened 2021-11-16
Update Mirage to latest PyPi version 2.2.0

jaytmiller:
Update/un-pin mirage version to pick up latest pypi (2.2.0 now) and rebuilt frozen requirements for jwebbinar.   Barring other inputs this is only recommended for merge after completion of the current jwebbinar.


cslocum (approved):
Looks good!

PR #176, opened 2021-11-29
Add check_cluster.py EKS and Hub health check script.

jaytmiller:
Added cluster_check.py to automate basic cluster and hub checks suitable for running automatically
  following push button or commit driven cluster and/or hub deployments,  or just as a quick look.
  Also includes one set of saved command results and a standalone test spec which is used to define
  and/or customize tests outside the script source code.
Added automatically defined JH_HOSTNAME to infrequent-env.
Modified image-pull to remove 'unscanned-' by default and/or permit specification of the IMAGE_ID to pull.

### DIFF
--- a/infrequent-env
+++ b/infrequent-env
@@ -5,6 +5,11 @@ export PATH="`pwd`/tools":${PATH}
 # ----------------- vvvv less frequently changed inputs vvvv -------------------------------
 #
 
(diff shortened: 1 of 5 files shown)


cslocum (approved):
This is fantastic!  Really nice work.  I can see this being really useful :)

PR #180, opened 2021-12-02
Update get-datadog-api-key

jmatuskey:
The current get-datadog-api-key gets extra text with the secret, this is also an improvement because it uses the query mechanism built in to awscli


cslocum (approved):
Sweet!  Thanks.

cslocum:
I forgot this was here and was about to PR my own fix.  So, I'm going to merge this instead...

PR #183, opened 2021-12-09
Cull unused tools scripts

jaytmiller:
Culled unused scripts from tools.   Simplified image-scan to reduce scripts.
Initial cleanup,  there's probably more we can delete.


.github/workflows/quality.yml
```diff
@@ -0,0 +1,54 @@
+name: Code Quality
+on:
+   schedule: # weekly
+     - cron: '0 0 * * 0'
+   push:
+      branches:
+        - main
+        - test
```
cslocum, line 8: Have we started using this branch scheme yet?

tools/efs-setup-crds
```diff
@@ -1,16 +0,0 @@
-#! /bin/bash
```
cslocum, line 1: Has this been made obsolete because we've added this to the helm configs?

tools/image-scan
```diff
@@ -1,7 +1,191 @@
...
+def get_scan_results(image_digest):
+    image_tag = os.environ["IMAGE_TAG"]
+    try:
+        return _get_scan_results(image_digest, image_tag)
+    except Exception:
+        print(
+            f"Failed fetching {image_tag}.  Trying for approved tag.", file=sys.stderr
+        )
+        if image_tag.startswith("unscanned-"):
+            return _get_scan_results(image_digest, image_tag[len("unscanned-") :])
```
cslocum, line 89: I must be missing something here.  How can we get results for an image that has not been scanned (to completion at least)?

cslocum (approved):
This is great Todd :)  Much cleaner!

I left a few questions, but none are blockers.

.github/workflows/quality.yml
jaytmiller: yeah.  push:main is essentiall merge PR.   push:test is a branch I sometimes create just to trigger tests on my fork to avoid messing with main.  (That said,  I'm inconsistent.)

tools/efs-setup-crds
jaytmiller: Jake has other methods.   I set this up in about 20 minutes for some webbinar at some point but (a) it's never been used for real and (b) I think it currently leaves files behind owned as root.   If we ever need it we can just back up in git.

tools/image-scan
jaytmiller: You're right,  it's possible this is just a logic error and ill-conceived  response to the error which started occurring when we started using unscanned-;  IMAGE_TAG basically became wrong in this context.

It's not out-of-the-question to drop the script altogether but IMHO it's still good to be able to fetch the results ourselves and correlate them to Ubuntu...  even though Ubuntu can never do anything about the results.

If you want to clean it up more post-merge,  have at it.

jaytmiller: One other possible thing:  if the scan fails,  it is never retagged.

cslocum: For sure.  It's definitely worth having around.  I think it's fine, I was just staring at it and wondering how that was supposed to be handled :+1:

PR #184, opened 2021-12-15
Check_cluster.py improvements

jaytmiller:
Fixed JH_HOSTNAME for jwebbinar dev.
Added pod log scanning
Added pod to node mapping
Added constants for NOTEBOOK_EC2_TYPE and MAX_RESTARTS.
Added builtin functions as source of check inputs.
Added rotated nodes only name check.
Misc improvements to error reporting,  dropped redundant "all pods" check
Turned off some formatting warnings from flake8 which made it past Black.
Added exception trapping for parser and assertion evaluation errors.
Added --exclude-groups.


cslocum (approved):
This is really great!  I'm liking this script a lot.

PR #185, opened 2021-12-16
Generalize get-datadog-key

cslocum:
Have the script accept either "api" or "app".

Update call to script in `deploy-datadog`.


jaytmiller (approved):
LGTM,  roll it out,  future generations can perfect get-datadog-key!

PR #190, opened 2022-01-16
Finalize security update release.

jaytmiller:
Drops pip switches --no-build-isolation and --use-deprecated=legacy-solver.
  --no-build-isolation is handled by env-src-install when needed.
  --use-deprecated=legacy-solver caused apparent version 0.0.0 problems and is
  not consistently needed.

Remove obsolete jupyterlab v2 labextension installs from the
  jupyter-lab-extension-setup script.
Add -g to explicit npm installs to keep them from winding up in $HOME
  where they are unused unless running standalone.

Added env-npm-audit and env-yarn-audit scripts to address npm security concerns
  with any easily available fixes.  These are a WIP due to substantial growth
  in image size and not active.

Re-compile Tike with tweaked versions.

Added list of failed packages to test-imports output.

Added import tests to jwebbinar.

Improved roman-cal import tests.

Dropped obsolete roman cubeviz package from roman-cal in favor of jdaviz.

Limit image-build-all to floating images;  purpose: update frozen requirements all missions.


jaytmiller:
Once the build tests pass,  this is ready for review.   It includes R&D code for updating vulnerable node packages,  but it is not active.

deployments/common/image/common-scripts/env-npm-audit
```diff
@@ -0,0 +1,38 @@
+#! /bin/bash
```
cslocum, line 1: Awesome!

deployments/common/image/common-scripts/jupyter-lab-extension-setup
```diff
@@ -1,30 +1,14 @@
 #! /bin/bash
 
-set -eu
+set -eux
 
 jupyter nbextension enable --py --sys-prefix  ipyvuetify
 jupyter nbextension enable --py --sys-prefix  ipysplitpanes
 
-jupyter labextension install \
```
cslocum, line 8: Did you discover that installing packages this way was causing problems?  (versus npm installs)

cslocum (approved):
This is really great work!  Nice job.

I left a few comments, none of them are blockers, mostly my own curiosity.

deployments/common/image/common-scripts/jupyter-lab-extension-setup
jaytmiller: I think those are all obsolete now due to JL  packaging improvements.   So e.g. for ipyvuetify the install instructions say:
`
$ pip install ipyvuetify
$ jupyter nbextension enable --py --sys-prefix ipyvuetify

# for Jupyter Lab < 3
$ jupyter labextension install jupyter-vuetify
`
and we are using jupyterlab 3+.   I think these days all the JL stuff happens in the pip install.

Likewise,  for the npm installs,  since the old installs did not say -g (global),  they were being installed in $HOME and not even visible on our hubs.   So I think they can all be removed,  although pip may be duplicating that work and will nominally still be there.

cslocum: Ah, thanks for that explanation, makes good sense.

jaytmiller:
The notebook tests for Tike are failing but I don't believe it is related to this build.    There are ".NET" failures and one of the setup actions done for image builds is to remove .NET to free up storage space.   For the sake of Jwebbinar I'm merging this now.

PR #191, opened 2022-01-18
Create new secrets decrypted "template" for Cognito RnD work

cslocum:
example-secrets-env-decrypted-cognito-rnd.yaml was added.

Note that this template also includes ACM related cert stuff, which will probably need to be replaced with our existing manual configuration as seen under the proxy section of https://github.com/spacetelescope/jupyterhub-deploy/blob/main/doc/example-secrets-env-decrypted.yaml.  I decided to leave this in here as it was the working example I used.


jaytmiller (approved):
Looks great!

PR #192, opened 2022-01-19
Doc for DD user metrics script

cslocum:
Addresses https://jira.stsci.edu/browse/JUSI-853

I wasn't sure what other information I should include here, so I'm definitely open to suggestions if you guys think of any other info that should be in the doc.

### DIFF
--- a/doc/DATADOG_USER_METRICS.md
+++ b/doc/DATADOG_USER_METRICS.md
@@ -1,0 +1,2 @@
+# Collecting user metrics from DataDog
+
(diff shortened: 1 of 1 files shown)


jaytmiller:
What I was looking for were Python-level comments which describe how you map particular metrics from DataDog to "user hours per week".     (So...  user hours per week is not a direct metric from DD,  or is it?  If it's not a direct metric,  how do you compute it?   Put that in the code.)

PR #195, opened 2022-01-31
Address scanning issues

jaytmiller:
Primarily for ECR but also DependaBot


cslocum (approved):
This looks great!  Thanks.

PR #203, opened 2022-03-23
git-sync blob-less clone

jaytmiller:
Added switch to git-sync clone to do a blobless clone.    This avoid downloading
every block in the repo history,  but does include the history and tree.   In theory
historical blobs are transparently downloaded if they are ever required.

This has two impacts due to speeding up post-startup-hook run times:  (1) faster logins (2) faster docker builds for the final steps which get repeated the most.

For tike cloning spacetelescope/notebooks,  2 minutes turns into 8 seconds.

We should hold this PR until (a) other updates are giving us clean builds/tests and (b) it won't introduce trouble for an imminent webbinar.


cslocum (approved):
Nice :)  I'm in favor of getting this in the main codebase now...if there are issues then we can work them out.

PR #204, opened 2022-03-28
Jusi 874 chilly

jmatuskey:


jmatuskey:
mistake

.github/workflows/update-frozen.yml
```diff
@@ -46,6 +48,12 @@ jobs:
            tools/image-configure ${{ matrix.DEPLOYMENT_NAME }}  0
            df -h
 
+      # - name: Set Chilly Build
```
jaytmiller, line 51: Does this need to be uncommented?

.github/workflows/update-frozen.yml
```diff
@@ -46,6 +48,12 @@ jobs:
...
+      #   shell: bash
+      #   run: |
+      #     source setup-env
```
jaytmiller, line 54: Once you do "source setup-env" you should have the option to just "cd $IMAGE_DIR".

tools/image-chilly-swap
```diff
@@ -0,0 +1,4 @@
+# fuzz out mission specific pip reqs only to create a build with less security issues
```
jaytmiller, line 1: Does this work w/o doing "git mv" vs "mv"?   Eventually a GitHub Action to PR the diffs will happen.  Potentially it should just blow away env-frozen-old unless it all works out because env-frozen was already in git and env-frozen-old was not.  In any case,  IMHO env-frozen-old should not be saved.

.github/workflows/update-frozen.yml
jmatuskey: removing this but it did NOT work for some reason

tools/image-chilly-swap
jmatuskey: This tool is like if you want to do a manual chilly build it just swaps chilly for frozen and wasn't intended to necessarily be related to git. It seemed analogous to how the frozen build gets made but not checked in every time you run image-build (as in this is not a GH action)

.github/workflows/update-frozen.yml
```diff
@@ -1,7 +1,9 @@
 name: Update Frozen Requirements
 on:
-   schedule:  # weekly
-     - cron: '0 0 * * 0'
+  schedule:  # weekly
```
jaytmiller, line 3: Before we close this,  I think we should modify this workflow so that every weekend GitHub does a chilly build.   That's essentially what this does now...  a floating build followed by automatically saving the resulting frozen requirements.  (What's missing is the chilly part.)   So potentially this should do a chilly swap to make the frozen requirements chilly... then do a frozen build instead of floating...  then make sure that even frozen builds re-freeze (which I believe I already changed but check).   Then the remainder of the script should save the new frozen requirements on a branch (should be close,  might be some git tweaks due to chilly swap).

jaytmiller (changes requested):
I wouldn't say "must be addressed",  I'd say "consider my suggestion before merging."

.github/workflows/update-frozen.yml
jmatuskey: sry just re-read, so you want to do another build to make sure that chilly still works

jaytmiller: For the current chilly,  my concept now is to (a) fuzz the frozen requirements and then (b) do a new frozen build to see how the fuzzed requirements are updated/overwritten.   (When I last messed with this though,  I was replacing the floating requirements and doing a floating build...  which then naturally produces new frozen requirements.   A thing to check for now is that there is no logic anywhere (in the common-scripts or tools) that prevents re-freezing during a frozen build...    which is now required as I understand it.    Once new frozen requirements are produced,  they are saved to a branch on github where they can be compared to the frozen requirements in main and/or tested.     If the updates turn out to be (a) useful and (b) harmless,  our ultimate goal is automatic weekly image updates on AWS for bugfix/security purposes only.

jaytmiller (approved):

PR #208, opened 2022-05-03
Remove mirisim

jaytmiller:
Remove the mirisim environment files and Dockerfile commands,  mainly to make unit tests pass again but also for cleanup. 

If mirisim needs to be restored roll back to a prior commit (e.g. just before this one) and take the old environment files and Dockerfile commands as an example which will be stale but hopefully otherwise useable.


cslocum (approved):
:+1:

PR #209, opened 2022-05-12
Chilly followup tweaks

jaytmiller:
Tweaked image-chill for OS-X portability.
Deleted Tike's env-chilly since it was triggering dependabot alerts and looked accidental.
Removed conda-tree and conda-build from common.conda and re-froze all missions;  this
   eliminates many dependencies from the pre-pip conda environment chilly does not cover.
Added chilly directories to .gitignore.


jmatuskey:
perl instead of sed for osx compatibility, that was a bit of a pain to make sure the backup file was explicitly not set when doing locally

jmatuskey (approved):

jaytmiller:
This PR is on hold while I straighten out issues with env-compile and chilly build.   Will PR again later once jwebbinar is updated for jdaviz meeting and this is reconciled with that.

PR #211, opened 2022-05-25
JWebbinar image updates for Jdaviz (DATB/INS) hack day

cslocum:
Update Ubuntu from Focal to Jammy
Update turboVNC from v2.2.6 to v3.0
Disable npm-audit; this was causing problems with labextensions but should probably be re-enabled at some point?

Note: this image/PR focuses on Jdaviz, but all of this is installed in the "JWebbinar" conda env/kernel


deployments/jwebbinar/image/environments/jwebbinar/jdaviz.pip
```diff
@@ -98,6 +99,17 @@ webencodings
 widgetsnbextension
 xlrd
 
+bqplot>=0.12.33
```
jaytmiller, line 102: Is there any reason not to just do these in common-env/jupyter.pip?   The idea there was to capture some of this difficult debug/setup for all the missions.   OTOH,  if that doesn't work,  maybe we should ditch common-env/jupyter.pip?  (Obviously you can do this though if it feels right,  pip-tools will let you have both,  and there are ways to omit jupyter.pip too.)

jaytmiller (approved):
Looks like a lot of great stuff,  couple Q's but go ahead and merge when you're ready.

hub/Chart.yaml
```diff
@@ -3,3 +3,5 @@ appVersion: '1.0'
 description: Deployment Chart for JupyterHub
 name: hub
 version: 0.0.1
+feature-gates:
```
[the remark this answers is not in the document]
cslocum: Hmm.  I don't know.  But the idea is that this won't be needed when we do update kubernetes.

deployments/jwebbinar/image/environments/jwebbinar/jdaviz.pip
cslocum: Good point.  I think I had just copied that from the jdaviz environment dir.  I'll give it a shot...

msanchezst (approved):
The image build tests are failing due to a dependency package not being available in pipy ( offending package: https://pypi.org/project/hsluv/). 

The changes in the PR are not causing the image build test failure.

deployments/jwebbinar/image/environments/jwebbinar/jdaviz.pip
cslocum: This did work.

PR #214, opened 2022-06-01
Final changes from JWebbinar/hack day (May 31)

cslocum:
I'll remove the "Download Imviz notebook; remove after May 31 2022" stuff shortly, but I'd like to get this merged in so we can tag this exact code.


msanchezst (approved):

deployments/jwebbinar/image/Dockerfile.custom
```diff
@@ -57,6 +57,11 @@ RUN   /opt/common-scripts/install-common  base
 
 # --------------------------- Jwebbinar -------------------------------------
 
+# Download Imviz notebook; remove after May 31 2022
+#RUN wget -O /home/jovyan/ImvizExample.ipynb https://raw.githubusercontent.com/spacetelescope/jdaviz/main/notebooks/ImvizExample.ipynb
+#COPY ImvizExample.ipynb /home/jovyan/ImvizExample.ipynb
+ADD https://raw.githubusercontent.com/spacetelescope/jdaviz/main/notebooks/ImvizExample.ipynb /home/jovyan/
```
jaytmiller, line 63: Based on the comment in line 60,  should the ADD be there?

jaytmiller (approved):
LGTM  but once we get this merged and tagged,   I'd like to do an update on main for the frozen and chilly requirements (which are about to get broken I'd guess).

jaytmiller:
It's interesting that the image builds didn't even run but the other workflows did.  They may be waiting for conflict resolution.

I have been thinking we should shift creation of Frozen requirements to an automatic step for a developer merge;   but one downside is that they don't always work and doing that might make it impossible to fix them.   still,  they lead to annoying conflicts like these all the time when devs do them so its something to think about.   Same with chilly requirements which now get generated at the same time floating requirements do.

deployments/jwebbinar/image/Dockerfile.custom
cslocum: I should have updated that comment.  ADD turned out to be the better way.

cslocum:
After reading your comments Todd, I'm sort of inclined to just close this PR.  There are conflicts - I do think that's why the image builds didn't run.  It's not super important that these changes get in.

deployments/jwebbinar/image/Dockerfile.custom
jaytmiller: Sorry,  I missed/forgot about your opening comment on the PR which mentioned this,  dumb question.

One thing did occur to me though:  does this fetch need to be in the post-startup-hook so it winds up in EFS instead of the image's  $HOME?

jaytmiller:
Closing like we discussed,  hopefully then last time we get burned on frozen requirements

PR #215, opened 2022-06-02
Added FREEZE_CHILL env var

jaytmiller:
Controls if/when new frozen and chilly requirements are copied out of
an image build into the git source tree,  nominally during CI/CD vs.
the floating developer builds.


cslocum (approved):
Awesome.  Thanks for this!

PR #216, opened 2022-06-03
Add version fuzzing to chilly spec for conda pkgs

jaytmiller:
Add requirements fuzzing to chilly conda env spec for conda pkgs
Add workflow_dispatch too all action workflows

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -3,12 +3,10 @@ on:
    push:
       branches:
         - main
-        - test
    pull_request:
       branches:
         - main
-        - test
-
+   workflow_dispatch:
 jobs:
   docker:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test.yml
```diff
@@ -3,12 +3,10 @@ on:
    push:
       branches:
         - main
-        - test
    pull_request:
       branches:
         - main
-        - test
-
+   workflow_dispatch:
```
cslocum, line 9: Why does `workflow_dispatch` have no attributes/values/config?

jaytmiller: Dunno,  just clone code Jake wrote to turn on the "run workflow" button elsewhere.

cslocum: OK, sounds good.

cslocum (approved):
:+1:

PR #217, opened 2022-06-13
 JWebbinar/PDRs4All 

msanchezst:
Preparation for upcoming Jwebinnar PDRs4All

Jwebbinnar deployment changes:
- Added Mirage data volume claims 
- Python 3.9 upgrade
- Pinned the required versions, including jwst with DMS build integration, mirage and photoutils

Common changes:
- Swapped Conda for Mamba 
- Upgraded JupyterHub to v1.5


jaytmiller (changes requested):
Nice!  There's a lot to unpack here.   (1) One superficial thing I did notice was that the import CI tests are not passing;  did a lot of packages get dropped?  If so,  maybe just move the environments/jwebbinar/tests/imports file to imports.disabled or something.   (2) Another more substantive thing I noticed was that the jwebbinar floating requirements are nailed down *hard*.  That's not necessarily wrong if they are an external input,  say given to us verbatim by some other group.   But if not,  the floating requirements should stay as whatever initial inputs you give to pip to derive that explicit list.  The *frozen* requirements will be an explicit list.   This also carries over to the chilly build,  where if the initial floating build is pinned hard,  so will chilly be.   It is not a mistake that exactly the floating requirements are in chilly too,  generally they're supposed to float and more restrictive requirements will come from ~= processed frozen requirements.  But where the floating requirements are pinned... chilly has to honor that too.   When all of them are pinned...  chilly will not budge unless there are unspecified dependencies.  For now,  I think we should probably only address (1).   I just wanted to clarify floating/frozen/chilly some if it was not already evident.   (Let me know if it's still unclear,  that would not be a shocker.  :-) )

msanchezst:
@jaytmiller  thanks for the review. 

(1) Yes, we removed the jdaviz package for this Jwebbinar.  My understanding is that most likely we will re-enable it after this jwebbinar, so it probably makes sense to disable it rather than removing the offending packages. The changes are already made.

(2) We had very specific package requirements from the team, so we had to build with python3.9 and specific versions of Jwst, mirage and photutils.

deployments/jwebbinar/image/Dockerfile.custom
```diff
@@ -43,6 +43,17 @@ RUN cd /tmp && \
     make && \
     make install
 
+# Download and unzip PDRs4All dataset from Box
```
jmatuskey, line 46: Is this a large amount of data? It might not be right to put this data in the image if so - better to pre-load it into EFS and get it mounted in the PVC

cslocum:
> Another more substantive thing I noticed was that the jwebbinar floating requirements are nailed down hard

Unfortunately, in this case, they were explicitly requested.  DMD release 1.4.6.  You can find the spec file in https://github.com/astroconda/astroconda-releases/tree/master/jwstdp/1.4.6.  The specs you see in the jwebbinar environment came directly from there (a literal copy/paste).

cslocum (approved):
This looks great.  Thanks!

jaytmiller (approved):
Agree this is great!   All the image build tests are passing,  the rationale for pinning the jwebbinar requirements hard is sound,  the one Safety issue conflicts with the exact pinning requirements which were given so reasonable arguments exist for fixing or not fixing.   Also given a new PVC, new data downloads, an update to common,  and the switch to mamba there's quite a variety of stuff here.   Nice work!

PR #219, opened 2022-06-27
Disabled data download scripts for PDRs4All JWebbinar.

msanchezst:
Since JWebbinar 15 is completed, this update disables the inline script to pull the notebook data at hub startup.


cslocum (approved):
Looks good!

PR #220, opened 2022-07-06
Comment out usage of Safety Python dependency scanner

jaytmiller:
due to revised licensing terms.


cslocum (approved):
OK.  Too bad the licensing changed.

PR #224, opened 2022-07-18
Improved chilly constraints

jaytmiller:
Improved chilly builds by limiting floating .pip carry-over to chilly builds to new build-hints.pip files which specify only requirements needed to help pip find a solution;   lots of .pip files formerly in env-chilly were deleted.  This enables heavily pinned DMS builds to chill normally since their .pip requirements are no longer included verbatim.

Factored chilling out into a post build step where it can be hacked on w/o rebuilding frozen requirements:  image-chill.   Tweaked the regexes used to chill to exclude versions which don't match x.y.z.  This prevents chilly from adding ~= to x.y only requirements and drifting in y,  or weird requirements like 2022a and maybe finding no ~= solution.

Got jwebbinar building again,  refroze and chilled.


msanchezst:
I think the update on this file is by mistake: deployments/jwebbinar/image/Dockerfile

The data pull was removed in the main branch a few commits back. Probably this branch was created prior to that change being made and it hasn't been updated since then.

msanchezst (approved):
Looks good to me! Thanks for this change, it seems like it will help with dependency hell

cslocum (approved):
Very nice!  This looks really useful.

jaytmiller:
Following up on @msanchezst 's comment about the Dockerfile,  it's a generated file (image-update).   So there may still be an issue with Dockerfile.custom or Dockerfile.trailer;  possibly Dockerfile was updated directly by mistake.

msanchezst:
@jaytmiller you are correct, that piece of code was added to the Dockerfile.custom in PR#217 and then removed from Dockerfile but not from Docker.custom in PR#219. 

I'll open a new PR to fix it.

PR #225, opened 2022-07-26
Removed PDRs4All Jwebinnar data download

msanchezst:
Removed PDRs4All Jwebinnar data download script from Dockerfile.custom and Dockerfile to speed up the notebook boot time.


jaytmiller (approved):
Looks great,  merge away

PR #230, opened 2022-08-18
JUSI-965 JH Helm Config Overwrite Checker

jaytmiller:
Attempts to improve on the situation where our 4 Helm config files for JH (all, common, sdlc, secrets) are combined with 
  silent replacements occurring potentially losing intended config.

For managing JH YAML config overwrites,  added Helm command to dump
  combined values of last JH Helm release to deploy-jupyterhub.

Also added scripts check-config and yaml-merge to automatically choose
  the YAML configs associated with the current mission,  merge them,
  and issue warnings when downstream files replace content from earlier
  files in the merge.   Run "check-config"

### DIFF
--- a/tools/check-config
+++ b/tools/check-config
@@ -1,0 +1,21 @@
+#! /bin/bash -eu
+
+# Script to detect possible config/values overwrites which normally happen silently in Helm.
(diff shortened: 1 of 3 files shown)


msanchezst (approved):
Very nice script!

PR #233, opened 2022-09-07
Add fuse fixes

msanchezst:
The s3 pods didn't get DNS resolution and couldn't connect to the S3 endpoints. There seems to be a bug with the Alpine image used in K8s which could be happening in this case. 

https://github.com/kubernetes/kubernetes/issues/30215

Changed the  base image to Ubuntu trusty (pinned) and latest versions of goofys and s3-fuse (also pinned)

Initial testing shows that the issue is corrected and it seems more stable than the previous version


jaytmiller (approved):
Lookd great,  thanks for seeing this through!

PR #235, opened 2022-09-12
JWebbinar updates for MAST event (not finalized)

cslocum:
Besides routine updates for JWebbinar, remove VNC and dependencies.


deployments/jwebbinar/config/common.yaml
```diff
@@ -21,7 +21,7 @@ jupyterhub:
         c.KubeSpawner.lifecycle_hooks = {
             "postStart": {
                 "exec": {
-                    "command": ["/bin/sh", "-c", "rsync -r /etc/default-home-contents/ /home/jovyan; /opt/environments/setup-notebooks; chmod +x /home/jovyan/Desktop/*.desktop; /opt/common-scripts/symlink-crds; /opt/environments/remove-kernel jdaviz"]
+                    "command": ["/bin/sh", "-c", "rsync -r /etc/default-home-contents/ /home/jovyan; /opt/environments/setup-notebooks; chmod +x /home/jovyan/Desktop/*.desktop; /opt/common-scripts/symlink-crds; /opt/environments/remove-kernel jdaviz; rm -rf /home/jovyan/.user-dirs/Desktop"]
```
cslocum, line 24: @jaytmiller - note the removal of `/home/jovyan/.user-dirs/Desktop` at the end.  This should be generalized somehow.  

I've been wondering if we should have a common command string or something that we run here, along with mission specific stuff...we've acquired all this cruft that is really only necessary to keep existing users' environments clean, and I'm betting there will be more as time goes on.

jaytmiller: I think all the cruft should be factored out into environments/post-start-hook and written as normal shell script(s).   That script should have two modes depending on whether or not the image is running on the hub;  In the common Dockerfile.trailer the script gets called with a "docker" parameter to set up the built-in $HOME for standalone runs,  in the config YAML it is called with no parameter for running on the hub with AWS services available.   Once the script is out of the list syntax into normal bash it's easier to deal with.    We could toss in a script of common things if needed which can be called by the mission specific post-start-hook scripts.

deployments/jwebbinar/image/environments/jwebbinar/build-hints.pip
```diff
@@ -1,2 +1,2 @@
-sphinx-asdf==0.1.3
-gwcs==0.18.0
\ No newline at end of file
+#sphinx-asdf
```
jaytmiller, line 1: Build hints are always "expendable",  but their purpose is to be carried over to chilly dependency resolution.   These are particularly strict or weird requirements w/o which chilly cannot find the overall solution since ~=x.y.z for these is not precise enough.  So #sphinx-asdf meaning"once used as a hint" makes sense,  but gwcs with no version should not help anything.

jaytmiller (changes requested):

deployments/common/image/common-env/common.pip
```diff
@@ -1,5 +1,5 @@
 papermill
-nbconvert
+#nbconvert
```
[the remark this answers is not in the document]
jaytmiller: Possible this has changed,  the real dependency for notebook testing is papermill.   I'll look tomorrow.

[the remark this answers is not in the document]
cslocum: This was causing a dependency conflict.  Agreed, it should go back in.

deployments/jwebbinar/image/environments/jwebbinar/build-hints.pip
cslocum: Like nbconvert, this should go back in.  It was causing conflicts.

jaytmiller:
Merging this with the understanding it is not yet finalized but will be before the webbinar.

PR #236, opened 2022-09-14
cleanup items

msanchezst:
Added cleanup items to tike dev


jmatuskey:
thanks! the cull might not need to be removed after all but it's not hurting anything either way (in values default)

PR #238, opened 2022-09-20
services in YAML don't overwrite existing services

jmatuskey:
in the extraconfig it removes existing services like the culler


jaytmiller (approved):
Much nicer,  looks great!

PR #240, opened 2022-09-29
Updated JH SSL cert on/from roman dev

jaytmiller:
curl was failing due to cert issues on roman dev.  using --insecure worked.  image-update found a new cert.


jaytmiller:
This did not solve my problem.  Closing

PR #241, opened 2022-10-02
Fix typo in env-conda from @explicit changes

jaytmiller:


cslocum (approved):
I don't see a change.  Is it just whitespace?  Either way, approved.

jaytmiller:
The fix is replacing a space with an underscore turning "env create" into a function call "env_create" which does the right thing for .explicit and .yaml

PR #243, opened 2022-10-07
Announcements R&D update

jaytmiller:
Overall adds an announcement service managed by JupyterHub and
  adapts EFS quota to output warnings and info to the service.
  User notebook sessions run an extension which periodically pulls
  messages from the announcement service.   A command line client
  "announce" supports sending, clearing, and fetching messages
  from a CI-node.
Added jupyter-resource-usage to common-env/jupyter.pip to display
  memory metrics from the REST API on the status bar at the bottom
Added klogs + kexec convenience bash functions to dump logs or
  run commands in/from pods with names matching a regex.
Updated SSL certs on/from roman dev, improved tools/cert-list
Added -f to deploy-jupyterhub chart removal in case deploy-jupyterhub is killed.


deployments/common/image/common-env/jupyter.pip
```diff
@@ -25,3 +25,5 @@ ipyvolume
 ipyvuetify
 ipywebrtc
 ipykernel
+jupyter-resource-usage
+git+https://github.com/jaytmiller/nersc-refresh-announcements@octarine-updates
```
msanchezst, line 29: Just checking that you want to include from your own fork

jaytmiller: Actually,  no,  nice catch!  But I fix it in the next PR.

msanchezst (commented):
This looks great!  I feel I don't know/understanding it enough to approve it but it looks fine to me. 

Thanks!!

msanchezst (approved):

PR #244, opened 2022-10-10
Updated tike build

jmatuskey:
able to reduce pins to only 4 - all in build-hints.pip. frozen and chilly updated as well 

Had to remove radvel from the binary installs which worked out great because with new dependencies it just pip installed

Need help verifying that this is correct!


deployments/tike/image/env-chilly/tess/build-hints.pip
```diff
@@ -1,18 +1,12 @@
...
-
-
-astroML>=1.0a1
-
-transitleastsquares>=1.0
+boto3==1.24.59
+jupyter-client<7.3.5
+mkl-fft<1.3.1
+jupyter-packaging<0.12
+# broken in everest look for a fix
```
jaytmiller, line 5: Nice comments!

jaytmiller (approved):
Looks great,  this was definitely a tough one!

PR #246, opened 2022-11-07
Announcement service protoype -> production

jaytmiller:
Updated announcements URL to https://github.com/spacetelescope/nersc-refresh-announcements 
Announce script python re-write
Added "replace" regex to announcements to suppress duplicate routine quota infos.


msanchezst (approved):
Looks great! Thanks!

PR #247, opened 2022-11-09
update helm chart version

jshihstsci:


msanchezst (approved):
Looks good to me. Last time I upgraded the hub version I think I only changed this file as well.

PR #248, opened 2022-11-15
OAuth and role/scoped perms for EFS quotas and announcement services

jaytmiller:
These are improvements to service security enabled by the update to JupyterHub-3.0
EFS quotas traded admin perms for scopes like those used by the idle culler.
Announcements relinquished admin perms altogether.
Authentication from notebook servers to the announcement service was converted to OAuth with confirmation (separate authorize page) turned off since it is a background service.
Calls to GET announcements using username "all" return a 404 since only "system" and real usernames work for GET.
Updated example secrets with new service configuration.

### DIFF
--- a/deployments/common/services/announcements.py
+++ b/deployments/common/services/announcements.py
@@ -5,10 +5,11 @@ import os
 import sys
 from collections import defaultdict
 import re
+import uuid
 
 from tornado import escape, ioloop, web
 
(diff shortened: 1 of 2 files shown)


msanchezst (commented):
Looks good to me. Thanks for the updates

jmatuskey (approved):
nice!

PR #250, opened 2022-11-23
Update for docker-stacks for new base image docker-stacks-foundation

jaytmiller:
Adds a new image between our cert image and base-image:  docker-stacks-foundation.  Normally docker-stacks-foundation now directly follows Ubuntu instead of base-image.
Modified image-build-all to build serially so it works with colima.


cslocum (approved):
Nice.  Thanks for doing this.

PR #251, opened 2022-11-28
Ditch .x file complexity

jaytmiller:
Removed code related to Docker build package download caches to simplify overall framework and get rid of .x file complexity.   AFAIK nobody uses it and it causes the "real" Dockerfile to be named Dockerfile.x,  involves generation code,  configuration code,  etc.


jaytmiller:
This will get done as part of partitioning jupyterhub-deploy into public and AWS repos

PR #254, opened 2022-12-29
Feature/idle culler fix

msanchezst:
Pinned sphinx package to ~=5.3.0  Latest version throwing compatibility errors with other libraries
Pinned glue-jupyter to ~=0.14.2 Latest version throwing compatibility errors with other libraries

Removed culling options for the hub-managed culler service. Culler is running on notebook servers now.

Cleaned up previous kernels from the last workshop

Swapped deprecated sklearn package with new scikit-learn package


deployments/common/image/common-env/docs.pip
```diff
@@ -1,5 +1,5 @@
 matplotlib
-sphinx
+sphinx~=5.3.0
```
jaytmiller, line 2: Wondering if we should pin in common-env or not since it's shared by all deployments;  alternative build-hints.pip.   Guess pinning in common-env does keep base + mission kernels consistent,  I'm just wondering if it's coming from a framework deficiency or is just outright the better choice.   The property of ONLY worrying about build-hints.pip to build each kernel makes them easier to maintain individually.   Seems like the minimum requirement for the pinning is keeping base + mission kernels consistent within one deployment,  not keeping all kernels of all deployments consistent which is more likely to be impossible.   No ironclad opinion personally,  just making sure we're consciously making this choice as "easiest overall."

infrequent-env
```diff
@@ -17,7 +17,8 @@ export JH_HOSTNAME=`echo ${JH_HOSTNAME} | sed -e's/tike/timeseries/g'`
 # Select "conda" tool used to install conda package and environments: conda or mamba
 #   mamba is largely a drop-in replacement for mamba with improved dependency resolution,
 #   parallel downloads, and C++ implementation.
-export CONDA_VER=mamba
+#export CONDA_VER=mamba
+export CONDA_VER=conda
```
jaytmiller, line 21: Not sure conda vs. mamba matters really,  but was this change intentional or accidental?  If accidental,  fine.  If intentional,  why?

jaytmiller (commented):
We should chat about a few choices,  but generally looks great,  choices are just choices not necessarily problems.

deployments/common/config/all.yaml
```diff
@@ -6,6 +6,10 @@ jupyterhub:
     extraConfig:
       spawning: |
         c.KubeSpawner.start_timeout = 15 * 60  # Up to 15 minutes
+        c.KubeSpawner.args.append('--NotebookApp.shutdown_no_activity_timeout=14400') # Kills notebook pod if no kernels or terminals opened after 4 hours
```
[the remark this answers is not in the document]
msanchezst: Good question. These were some numbers chosen more for testing than for production.

The reasoning behind the 1 hour to kill an idle kernel is to avoid memory loss when running a notebook and the user doesn't provide any input for 1 hour. I was thinking that giving 1 hour to kill an idle kernel may actually be low...Users may take a break or need to do some other research with other tools while using the notebooks... 

Regarding the notebook shutdown, I agree, 4 hours seems excessive. Maybe we can set the killing of idle kernels to 2 or 3 hours, and then shutdown the notebook server 30 minutes after no kernels. 

About the terminals opened, this will prevent culling. There seem to be options to also allow for culling with terminals opened, but that hasn't been tested yet.

infrequent-env
msanchezst: I was having dependency issues while building the images with mamba, swapped the conda, and seemed to work fine. 
My understanding is that it doesn't really matter which one we choose as far as it works.

deployments/common/image/common-env/docs.pip
msanchezst: Good point. This was an interesting case due to some dependency resolution pointing to alpha versions of other packages, which I assume will be resolved during the next package updates.  

After reading your comments I believe it makes more sense to only do it at the environment level, I will be updating the PR, Thanks!

infrequent-env
jaytmiller: I think they're very similar,  but may actually have different package repos,  conda-forge vs...  something else.   I think docker-stacks is now using mamba which is one consideration...  but I think our switch to mamba was pretty arbitrary so there's no huge reason not to switch back.

deployments/common/config/all.yaml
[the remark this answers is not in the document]
jaytmiller: I'll add a little ticket to discuss and/or add terminal culling.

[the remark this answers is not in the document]
msanchezst: I temporarily tweaked the values to:  

- 4 hours with no kernel activity kills the kernel
- 30 mins with no kernels or terminals running kills the notebook pod.

deployments/common/image/common-env/docs.pip
msanchezst: Removed the pinning from common

infrequent-env
msanchezst: Swapped back to mamba

deployments/common/image/common-env/jupyter.pip
```diff
@@ -2,7 +2,7 @@ jupyterhub
 jupyterlab
 notebook
 bqplot
-glue-jupyter
+glue-jupyter~=0.14.2
```
jaytmiller, line 5: Move this to build-hints.pip too

msanchezst: Just did it on the last commit ce85d3f

jaytmiller: perfect

infrequent-env
msanchezst: Since mamba was throwing errors, swapped back to conda. Will investigate the issue.

jaytmiller: Manuel switched back to conda and was able to build on a CI-node.  jwebbinar failed GH Actions for an unrelated reason so we're pressing ahead with the merge and transition to codecommit before further debug.

jaytmiller (approved):
Overall looking great!   (Freak jwebbinar build failure to be debugged later,  they're endless)

PR #256, opened 2023-01-06
Update roman 

jaytmiller:
This addresses the broken Roman DEV floating builds by:

1. Dropping obsolete build fixes / hints
2. Updating the roman-cal saved requirements to their 0.9.0 versions from artifactory,  mainly to get a more viable .explicit conda environment.
3. Adding new build hints to resolve new conflicts.

The dropped file **git-clone-filter-blob-none-fix.conda** in particular was a crude "build hint" which (a) was a hint for the basic roman-cal conda environment and (b) didn't carry over to frozen or chilly builds like normal build-hints.pip would.   This obsolete hint was resulting in conflicts between pip-tools and pip.


deployments/common/image/common-scripts/env-conda
```diff
@@ -33,6 +33,7 @@ env_spec=/opt/environments/${env}/${env}.@(yml|explicit)
 
 env_frozen_yml=/opt/env-frozen/${env}/requirements.yml
 env_chilly_yml=/opt/env-chilly/${env}/requirements.yml
+hints=/opt/environments/${env}/buld-hints.conda
```
msanchezst, line 36: Looks like there is a typo in the filename (build vs buld)

It doesn't matter since all the yml files are used but just letting you know.

msanchezst (approved):
Looks good to me!

deployments/common/image/common-scripts/env-conda
jaytmiller: That's a great catch!  It would break the intended behavior of having a "conda level" build-hints file.   Except for the typing impaired.  :-)

PR #257, opened 2023-01-09
Updated datadog deployment scripts to use tags from ENV VARS

msanchezst:
Updated datadog deployment script to use tags from ENV VARS and to pull specific helm chart version


jaytmiller (approved):
Looks great,  thanks!

PR #258, opened 2023-01-10
Replacing env var prod to ops to match datadog tags during dd agent

msanchezst:


jaytmiller (approved):
LGTM assuming you're ready

